PR #2, opened 2025-03-27
hono api example

abbaspour:
By submitting a PR to this repository, you agree to the terms within the [Auth0 Code of Conduct](https://github.com/auth0/open-source-template/blob/master/CODE-OF-CONDUCT.md). Please see the [contributing guidelines](https://github.com/auth0/.github/blob/master/CONTRIBUTING.md) for how to create and submit a high-quality PR for this repo.

### Description

> Describe the purpose of this PR along with any background information and the impacts of the proposed change. For the benefit of the community, please do not assume prior context.
>
> Provide details that support your chosen implementation, including: breaking changes, alternatives considered, changes to the API, etc.
>
> If the UI is being changed, please provide screenshots.

### References

> Include any links supporting this change such as a:
>
> - GitHub Issue/PR number addressed or fixed
> - Auth0 Community post
> - StackOverflow post
> - Support forum thread
> - Related pull requests/issues from other repos
>
> If there are no references, simply delete this section.

### Testing

> Describe how this can be tested by reviewers. Be specific about anything not tested and reasons why. If this library has unit and/or integration testing, tests should be added for new functionality and existing tests should complete without errors.
>
> Please include any manual steps for testing end-to-end or functionality not covered by unit/integration tests.
>
> Also include details of the environment this PR was developed in (language/platform/browser version).

- [ ] This change adds test coverage for new/changed/fixed functionality

### Checklist

- [ ] The correct base branch is being used, if not the default branch


examples/example-hono-api/README.md
```diff
@@ -0,0 +1,39 @@
...
+npm run deploy
+``` 
+
+## Endpoints
+
+The example API has the following endpoints:
+
+- `GET /api/public`: A public endpoint that can be accessed without authentication.
+- `GET /api/private`: A private endpoint that can only be accessed by authenticated users.
+- `GET /api/private/scope`: A private endpoint that can only be accessed by authenticated users with the `read:data` scope.
```
frederikprijck, line 37: nitpicking here, but can we use the same scope in all examples? The other samples are using `read:private`. If we do not like that scope, we can rename the others to `read:data` (I can do that, no need to do that in this PR, but curious to your thoughts?)

examples/example-hono-api/package.json
```diff
@@ -0,0 +1,16 @@
...
+  "type": "module",
+  "scripts": {
+    "deploy": "wrangler deploy",
+    "dev": "wrangler dev",
+    "cf-typegen": "wrangler types"
+  },
+  "devDependencies": {
+    "@auth0/auth0-api-js": "*",
+    "@cloudflare/workers-types": "^4.20250321.0",
+    "hono": "^4.7.5"
```
frederikprijck, line 14: Is this expected as a devDependencies? I assume this is fine because of how hono works, but just checking.

examples/example-hono-api/src/auth0.ts
```diff
@@ -0,0 +1,175 @@
...
+      });
+    }
+
+    const accessToken = parts[1];
+    if (!accessToken || accessToken.length === 0) {
+      const errDescription = 'No token included in request';
+      throw new HTTPException(401, {
+        message: errDescription,
+        res: unauthorizedResponse({
+          ctx,
```
frederikprijck, line 90: Do we need all of this for a sample application? I understand that, if we want to make it an actual SDK, we want to be more defensive and have clear errors.

But curious why we cant use what we use in other samples here?

```ts
const accessToken = ctx.req.raw.headers.get('Authorization')?.split(' ')[1];

      if (!accessToken) {
        return replyWithError(
          reply,
          400,
          'invalid_request',
          'No Authorization provided'
        );
      }
```

I feel like the code above unneccesary clutters the sample app, making it more complicated to parse.

examples/example-hono-api/src/auth0.ts
```diff
@@ -0,0 +1,175 @@
...
+          error: 'invalid_request',
+          errDescription,
+        }),
+      });
+    }
+
+    try {
+      const jwtPayload = await apiClient.verifyAccessToken({ accessToken });
+
+      ctx.set('jwtPayload', jwtPayload);
```
frederikprijck, line 100: All samples and SDKs set `user`, which I understand may actualy not always be correct.

However, I strongly believe we should be consistent.

I think `jwyPayload` isnt the best name for the users, should we use something different, but consistent (I can then update the other places, as well as the SDKs)

examples/example-hono-api/src/auth0.ts
```diff
@@ -0,0 +1,175 @@
...
+  errDescription: string;
+  statusText?: string;
+}) {
+  return new Response('Unauthorized', {
+    status: 401,
+    statusText: opts.statusText,
+    headers: {
+      'WWW-Authenticate': `Bearer realm="${opts.ctx.req.url}",error="${opts.error}",error_description="${opts.errDescription}"`,
+    },
+  });
```
frederikprijck, line 133: Any reason we arent returning it in the body? I know the `WWW-Authenticate` headers is what we need for spec compliance, but users typically prefer reading the body.

In the sample apps, I have always omitted the `WWW-Authenticate`, and only return the body.
In the SDKs, I ensured we have both (see [here](https://github.com/auth0/auth0-fastify/blob/early-access/packages/auth0-fastify-api/src/index.ts#L68-L78)).

This mostly to keep the sample apps easy to parse.

examples/example-hono-api/src/auth0.ts
```diff
@@ -0,0 +1,175 @@
...
+    const payload = ctx.var.jwtPayload as JWTPayload;
+
+    if (!payload.scope) {
+      throw new HTTPException(403, {
+        message: 'Forbidden',
+        res: unauthorizedResponse({
+          ctx,
+          error: 'insufficient_scope',
+          errDescription: `Missing required scope: ${scope}`,
+        }),
```
frederikprijck, line 148: As far as I know, `insufficient_scope` shouldnt be a 401 , any reason we are doing this here?

frederikprijck: Additionally, I feel like this is a lot of code to validate the scopes (we even throw the same error twice), while we use this in all other samples and SDKs:

```
function validateScopes(token: Token, requiredScopes: string[]): boolean {
  let tokenScopes: string[] = [];

  if (token.scope) {
    tokenScopes =
      typeof token.scope === 'string' ? token.scope.split(' ') : token.scope;
  }

  return requiredScopes.every((required) => tokenScopes.includes(required));
}
```

I feel like using this could simplify things.

frederikprijck (changes requested):
Left some comments, can we be consistent in terms of casing and semi colons usage ?

examples/example-hono-api/.env.example
```diff
@@ -0,0 +1,2 @@
+AUTH0_DOMAIN=
+AUTH0_AUDIENCE=
```
frederikprijck, line 2: Is it expected to have both this `.env.example` and `.dev.vars` file?

examples/example-hono-api/src/auth0.ts
```diff
@@ -0,0 +1,166 @@
...
+                throw new Error('Auth0 middleware requires options "auth0_domain"');
+            }
+            if (!auth0_audience || auth0_audience.length === 0) {
+                throw new Error('Auth0 middleware requires options "auth0_audience"');
+            }
+
+            apiClient = new ApiClient({
+                domain: auth0_domain,
+                audience: auth0_audience,
+            });
```
frederikprijck, line 42: Any reason not to stick to avoiding snake casing here and go with the below?

```suggestion
            const { domain, audience } = options || {
                domain: auth0Env.AUTH0_DOMAIN,
                audience: auth0Env.AUTH0_AUDIENCE,
            };
            if (!domain || domain.length === 0) {
                throw new Error('Auth0 middleware requires options "domain"');
            }
            if (!audience || audience.length === 0) {
                throw new Error('Auth0 middleware requires options "audience"');
            }

            apiClient = new ApiClient({
                domain,
                audience,
            });
```

If we do this, we should also change line 16 to 17.

examples/example-hono-api/src/auth0.ts
```diff
@@ -0,0 +1,166 @@
...
+export type JwtEnv = {
+    Variables: {
+        user: JWTPayload
+    }
+};
+
+export const jwt = (
+    options?: {
+        auth0_domain: string;
+        auth0_audience: string;
```
frederikprijck, line 17:
```suggestion
        domain: string;
        audience: string;
```

examples/example-hono-api/src/index.ts
```diff
@@ -0,0 +1,23 @@
...
+
+app.use('/api/private/*', jwt());
+
+app.get('/api/private/', (c) => {
+    return c.text(`hello ${c.get('user').sub}`);
+})
+
+app.get('/api/private/scope', requireScope('read:data'), (c) => {
+    return c.text(`hello ${c.get('user').sub}`);
+})
```
frederikprijck, line 20: Is there a specific reason to prefer splitting it in two functions:

- jwt
- requireScope

I feel like that is easy to make mistakes and forget to add jwt etc. Why not just have one, with an optional scope:

```suggestion

app.get('/api/private/', jwt(), (c) => {
    return c.text(`hello ${c.get('user').sub}`);
});

app.get('/api/private/scope', jwt({ scopes: ['read:data'] }), (c) => {
    return c.text(`hello ${c.get('user').sub}`);
});
```

But this will require internal changes as well, but it would align with our fastify-api SDK.

examples/example-hono-api/src/index.ts
```diff
@@ -0,0 +1,23 @@
+import {Hono} from 'hono';
+import {JwtEnv, jwt, requireScope} from './auth0';
+import { logger } from 'hono/logger'
+
+const app = new Hono<JwtEnv>();
+app.use(logger())
+
+app.get('/api/public', (c) => {
+    return c.text(`Hello world!`);
+})
```
frederikprijck, line 10:
```suggestion
app.use(logger());

app.get('/api/public', (c) => {
    return c.text(`Hello world!`);
});
```

examples/example-hono-api/src/index.ts
```diff
@@ -0,0 +1,23 @@
+import {Hono} from 'hono';
+import {JwtEnv, jwt, requireScope} from './auth0';
+import { logger } from 'hono/logger'
```
frederikprijck, line 3:
```suggestion
import { logger } from 'hono/logger';
```

examples/example-hono-api/src/auth0.ts
```diff
@@ -0,0 +1,166 @@
...
+    statusText?: string;
+}) {
+    return new Response('Unauthorized', {
+        status: 401,
+        statusText: opts.statusText,
+        headers: {
+            'WWW-Authenticate': `Bearer realm="${opts.ctx.req.url}",error="${opts.error}",error_description="${opts.errDescription}"`,
+        },
+    });
+}
```
frederikprijck, line 134:
```suggestion
function unauthorizedResponse(opts: {
    ctx: Context;
    error: string;
    errorDescription: string;
    statusText?: string;
}) {
    return new Response('Unauthorized', {
        status: 401,
        statusText: opts.statusText,
        headers: {
            'WWW-Authenticate': `Bearer realm="${opts.ctx.req.url}",error="${opts.error}",error_description="${opts.errorDescription}"`,
        },
    });
}
```

I think using `err` over `error` isn't helpful here, even mopre so because line 123 already mentions `error`.

frederikprijck (changes requested):

PR #20, opened 2025-03-31
fix(auth0-server-js): Do not document bindingMessage as optional

frederikprijck:
As our CIBA implementation follows the [FAPI CIBA spec](https://openid.net/specs/openid-financial-api-ciba.html), `bindingMessage` is not optional.

This was taken into consideration in the code, but not in the documentation.

This PR reflects this in the docs for auth0-server-js as well.


arpit-jn (approved):
LGTM

PR #22, opened 2025-04-03
docs(auth0-server-js): Update TransactionStore guidance to use AbstractTransactionStore

frederikprijck:
Even though the current guidance works, using the abstract `AbstractTransactionStore` instead of the `TransactionStore` interface would enable users to easily add encryption.


arpit-jn (approved):
LGTM

PR #23, opened 2025-04-03
chore: Use AbstractTransactionStore in sample applications

frederikprijck:
Even though the current example apps work fine, using the abstract AbstractTransactionStore instead of the TransactionStore interface enables us to use encryption.


arpit-jn (approved):
LGTM

PR #55, opened 2025-09-07
feat(auth0-auth-js): expose more granular CIBA methods

guabu:
### Description

Exposes more granular methods for the developer to have more control over the CIBA flow.

Currently, the `backchannelAuthentication` method takes care of the entire flow, including polling. The caller might want to have more control over the polling logic — in this case the flow is broken out into 2 separate methods: `initiateBackchannelAuthentication` and `backchannelAuthenticationGrant`.

### Testing

### Checklist

### DIFF
--- a/packages/auth0-auth-js/EXAMPLES.md
+++ b/packages/auth0-auth-js/EXAMPLES.md
@@ -328,6 +328,10 @@ const { authReqId, expiresIn, interval } = await authClient.initiateBackchannelA
 const tokenResponse = await authClient.backchannelAuthenticationGrant({ authReqId });
 ```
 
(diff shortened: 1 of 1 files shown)


packages/auth0-auth-js/EXAMPLES.md
```diff
@@ -314,6 +314,20 @@ const tokenResponse = await authClient.backchannelAuthentication({
...
+  bindingMessage: '',
+  loginHint: {
+    sub: 'auth0|123456789'
+  }
+});
+
+// Poll the token endpoint using the authReqId
+const tokenResponse = await authClient.backchannelAuthenticationGrant({ authReqId });
+```
+
```
frederikprijck, line 330: Can we add something in the md file here that explains what interval should be used for ? and what expiresIn does ?

PR #56, opened 2025-09-07
feat(auth0-server-js): return token set from CIBA flow

guabu:
### Description

Returns the token response from the CIBA flow to the caller. This is useful in cases where the CIBA flow happens out of band, such as when an AI agent acts on behalf of a user and a session may not exist.

The method is exposed as `getTokenByBackchannelAuth`.

### Testing

### Checklist

### DIFF
--- a/packages/auth0-server-js/EXAMPLES.md
+++ b/packages/auth0-server-js/EXAMPLES.md
@@ -563,6 +563,17 @@ await serverClient.loginBackchannel({
 });
 ```
 
+To obtain the tokens directly without establishing or updating a session, you can use the `getTokenByBackchannelAuth()` method. This is useful when CIBA is used out-of-band such as authorizing an AI agent to act on behalf of a user. The method accepts the same parameters as `loginBackchannel()` and returns the token response directly:
+
+```ts
+const tokenResponse = await serverClient.getTokenByBackchannelAuth({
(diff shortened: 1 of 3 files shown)


packages/auth0-server-js/EXAMPLES.md
```diff
@@ -563,6 +563,17 @@ await serverClient.loginBackchannel({
...
+
+```ts
+const tokenResponse = await serverClient.getTokenByBackchannelAuth({
+  bindingMessage: '',
+  loginHint: {
+    sub: 'auth0|123456789'
+  }
+});
+```
+
```
frederikprijck, line 576: Can we add a caveat message here that this will not store the token? As this can be confusing as all other getToken methods do store the tokens.

packages/auth0-server-js/src/server-client.ts
```diff
@@ -295,6 +296,28 @@ export class ServerClient<TStoreOptions = unknown> {
     };
   }
 
+  /**
+   * Performs a Client-Initiated Backchannel Authentication and returns the tokens.
```
frederikprijck, line 300: Can we add a caveat message here that this will not store the token? As this can be confusing as all other getToken methods do store the tokens.

frederikprijck (changes requested):

PR #68, opened 2025-09-18
feat(auth0-api-js): Add support for Token Vault to exchange access tokens

guabu:
### Description

Adds a `getAccessTokenForConnection` method to exchange an access token (`urn:ietf:params:oauth:token-type:access_token`) for a connection's access token using the `urn:auth0:params:oauth:grant-type:token-exchange:federated-connection-access-token` grant.

The API client can now optionally be instantiated by providing a client ID and secret to facilitate the exchange.

### Testing

1. Authenticate in a web app and obtain the access token for an audience
2. Pass the access token `getAccessTokenForConnection` method
3. The token set returned should contain the access token for the connection

### Checklist

### DIFF
--- a/packages/auth0-api-js/EXAMPLES.md
+++ b/packages/auth0-api-js/EXAMPLES.md
@@ -1,0 +1,25 @@
+# Examples
+
+- [Get an access token for a connection](#get-an-access-token-for-a-connection)
+
+## Get an access token for a connection
+
+The `getAccessTokenForConnection` method allows you to exchange an access token for an access token for a specific connection. To use this method, you will need to instantiate the `ApiClient` with the client credentials:
+```ts
+import { ApiClient } from '@auth0/auth0-api-js';
+
+const apiClient = new ApiClient({
+  domain: '<AUTH0_DOMAIN>',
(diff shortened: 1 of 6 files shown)


packages/auth0-api-js/EXAMPLES.md
```diff
@@ -0,0 +1,25 @@
+# Examples
+
+- [Get an access token for a connection](#get-an-access-token-for-a-connection)
+
+## Get an access token for a connection
```
pmalouin, line 5: There should be more detailed docs around each of the parameters. Use the [existing docs](https://github.com/guabu/auth0-auth-js/blob/main/packages/auth0-auth-js/EXAMPLES.md#retrieving-a-token-for-a-connection) as a reference.
image

pmalouin: Also, we are not describing here the shape of the `tokenSet` variable (ie. the return type of the method). I'd recommend that we somehow show its shape, at the very least, that the AT is available at `tokenSet.accessToken`.

pmalouin: non blocking ^

pmalouin (approved):

PR #165, opened 2026-04-15
fix(auth0-auth-js): derive challenge type from authenticator_type and oob_channel

GuaiamumSuspeito:
### Description

`mfa.getAuthenticators()` in `@auth0/auth0-spa-js` always returns an empty array, even when the user has active MFA authenticators enrolled.

The bug spans two packages:

**Layer 1 — `@auth0/auth0-auth-js` (this fix):** `transformAuthenticatorResponse` in `utils.ts` maps `api.type` to the response, but the Auth0 `GET /mfa/authenticators` endpoint does not return a `type` field. Per the [official API docs](https://auth0.com/docs/secure/multi-factor-authentication/manage-mfa-auth0-apis/manage-authenticator-factors-mfa-api#list-authenticators), the response uses `authenticator_type` and `oob_channel`:

```json
[
  {"authenticator_type": "recovery-code", "id": "recovery-code|dev_...", "active": true},
  {"authenticator_type": "otp", "id": "totp|dev_...", "active": true},
  {"authenticator_type": "oob", "oob_channel": "sms", "id": "sms|dev_...", "name": "+1123XXXXX", "active": true}
]
```

Since `api.type` is always `undefined`, every transformed authenticator has `type: undefined`.

**Layer 2 — `@auth0/auth0-spa-js` (downstream):** `MfaApiClient.getAuthenticators` filters authenticators with `if (!auth.type) return false`, which rejects everything because `type` is always `undefined`.

**The fix** adds a `deriveChallengeType(authenticatorType, oobChannel)` function that maps the API fields to the challenge types `auth0-spa-js` expects:

| `authenticator_type` | `oob_channel` | Derived `type` |
|---|---|---|
| `otp` | — | `otp` |
| `recovery-code` | — | `recovery-code` |
| `oob` | `sms` | `phone` |
| `oob` | `voice` | `phone` |
| `oob` | `auth0` | `push-notification` |
| `oob` | `email` | `email` |

Also adds `oob_channel` (singular) to `AuthenticatorApiResponse` and exports a new `ChallengeType` union type.

### References

- Auth0 MFA Authenticators API docs: https://auth0.com/docs/secure/multi-factor-authentication/manage-mfa-auth0-apis/manage-authenticator-factors-mfa-api#list-authenticators
- `@auth0/auth0-spa-js` `ChallengeType` definition: `src/mfa/types.ts`
- `@auth0/auth0-spa-js` filter in `MfaApiClient.getAuthenticators`: `src/mfa/MfaApiClient.ts`

### Testing

Added 8 unit tests for `deriveChallengeType` covering all authenticator type / OOB channel combinations and edge cases (unknown types, missing channel). Updated existing `listAuthenticators` tests to assert the derived `type` values (`'otp'`, `'phone'`) instead of `undefined`. Updated mock data to include `oob_channel` to match real API response shapes.

All 146 existing tests pass.

### Checklist

- [x] All active GitHub checks for tests, formatting, and security are passing

### DIFF
--- a/packages/auth0-auth-js/src/mfa/types.ts
+++ b/packages/auth0-auth-js/src/mfa/types.ts
@@ -30,6 +30,12 @@ export type AuthenticatorType = 'otp' | 'oob' | 'recovery-code';
 export type OobChannel = 'sms' | 'voice' | 'auth0' | 'email';
 
 /**
+ * Challenge types derived from authenticator_type and oob_channel.
+ * Used by downstream consumers (e.g. auth0-spa-js) to filter authenticators.
+ */
+export type ChallengeType = 'otp' | 'recovery-code' | 'phone' | 'push-notification' | 'email';
+
+/**
  * Represents an MFA authenticator enrolled by a user.
  */
 export interface AuthenticatorResponse {
@@ -43,8 +49,8 @@ export interface AuthenticatorResponse {
   name?: string;
   /** Delivery channels for OOB authenticators (only present for authenticatorType: 'oob') */
   oobChannels?: OobChannel[];
-  /** Additional type information */
-  type?: string;
+  /** Challenge type derived from authenticator_type and oob_channel */
+  type?: ChallengeType;
 }
 
 /**
@@ -187,7 +193,7 @@ export interface AuthenticatorApiResponse {
   active: boolean;
   name?: string;
   oob_channels?: OobChannel[];
-  type?: string;
+  oob_channel?: OobChannel;
 }
 
 /**
--- a/packages/auth0-auth-js/src/mfa/types.ts
+++ b/packages/auth0-auth-js/src/mfa/types.ts
@@ -31,7 +31,6 @@ export type OobChannel = 'sms' | 'voice' | 'auth0' | 'email';
 
 /**
  * Challenge types derived from authenticator_type and oob_channel.
- * Used by downstream consumers (e.g. auth0-spa-js) to filter authenticators.
  */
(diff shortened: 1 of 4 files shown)


packages/auth0-auth-js/src/mfa/types.ts
```diff
@@ -29,6 +29,12 @@ export type AuthenticatorType = 'otp' | 'oob' | 'recovery-code';
  */
 export type OobChannel = 'sms' | 'voice' | 'auth0' | 'email';
 
+/**
+ * Challenge types derived from authenticator_type and oob_channel.
+ * Used by downstream consumers (e.g. auth0-spa-js) to filter authenticators.
```
GuaiamumSuspeito: remove this, no guessing about downstream consumption

GuaiamumSuspeito (commented):
self-review

packages/auth0-auth-js/src/mfa/types.ts
```diff
@@ -43,8 +48,8 @@ export interface AuthenticatorResponse {
   name?: string;
   /** Delivery channels for OOB authenticators (only present for authenticatorType: 'oob') */
   oobChannels?: OobChannel[];
-  /** Additional type information */
-  type?: string;
+  /** Challenge type derived from authenticator_type and oob_channel */
+  type?: ChallengeType;
```
GuaiamumSuspeito, line 52: This is a library API break, not sure what's the approach on API breaks for the repo, but if you'd rather avoid it than doing a major bump or breaking consumers, we can keep it as a plain string and leave the conversion to @auth0/auth0-spa-js

PR #167, opened 2026-04-28
feat(auth0-api-js): add OBO (on-behalf-of-token-exchange) support 

nandan-bhat:
### Changes
Implemented `On-Behalf-Of` (OBO) token exchange support in `@auth0/auth0-api-js` by adding `ApiClient#getTokenOnBehalfOf(...)` as a thin wrapper over the existing `RFC 8693` exchange flow. 

The new API keeps the application-facing contract simple by accepting only the incoming access token, `downstream audience`, and `optional scope`, while reusing the existing confidential-client auth and token exchange behavior internally.

This PR also adds 
- Exported `OBO` result/types
- Test coverage for the new method
- `README.md` and `EXAMPLES.md` updates

No existing public APIs were changed, so this is additive and non-breaking.


kishore7snehil (approved):
LGTM!

PR #164, opened 2026-04-15
feat(auth0-server-js): add multi-factor authentication (MFA) support

amitsingh05667:
This **multi-factor authentication (MFA) support** is split into two PR's as mentioned below -
- **auth0-auth-js** - https://github.com/auth0/auth0-auth-js/pull/179
- **auth0-server-js** - https://github.com/auth0/auth0-auth-js/pull/164 (the current one)

## Summary
- Introduce `ServerMfaClient` module with comprehensive Multi-Factor Authentication (MFA) support for server-side flows
- Add four public methods: `listAuthenticators()`, `enrollAuthenticator()`, `challengeAuthenticator()`, and `verify()` (verifies MFA challenge and saves tokens into session)
- Integrate `ServerMfaClient` as `serverClient.mfa` property on `ServerClient`

## Design decisions
- **New feature only**: Does not introduce breaking changes to existing functionality
- **Dedicated error classes**: Specific error classes for each MFA operation (enrollment, listing, challenge, verification) for granular error handling
- **Server-side only**: MFA `verify()` automatically saves tokens into the user's session after successful verification, keeping credentials server-side
- **Delegates to `auth0-auth-js`**: `listAuthenticators`, `enrollAuthenticator`, and `challengeAuthenticator` delegate directly to the underlying `AuthClient.mfa`, keeping transport logic in one place

## Test plan
- [x] 28 unit tests covering all 4 operations in `server-mfa-client.spec.ts`
- [x] 228 total tests passing across 9 test files
- [x] Error path coverage
- [x] `npm run build` passes
- [x] Tested end-to-end against a real tenant

PR #169, opened 2026-05-07
feat: expose MFA context on token error `cause` with `isMfaRequiredError` type guard

subhankarmaiti:
## Summary         
 - Extends `OAuth2Error` interface with optional `mfa_token` and `mfa_requirements` fields                                                                              
  - Adds `MfaRequirements` interface describing which factors the user must challenge or enroll
  - Adds `isMfaRequiredError(error)` type guard that narrows any caught error to one with `cause.mfa_token` (string) and `cause.mfa_requirements` guaranteed present     
  - Propagates MFA fields from the server response body into the `cause` of existing error classes via a `toOAuth2Error` helper                                          
  - Updates EXAMPLES.md with documentation and usage examples                                                                                                            
                                                                                                                                                                         
  ## Motivation                                                                                                                                                          
                                                                                                                                                                         
  When the Auth0 server requires MFA, it returns a 403 with `error: "mfa_required"`, `mfa_token`, and optionally `mfa_requirements`. Previously these fields were        
  discarded when the response was wrapped into `TokenByPasswordError`, `TokenByRefreshTokenError`, etc.                                                                
                                                                                                                                                                         
  This change is **non-breaking**:                                                                                                                                       
  - Existing `instanceof TokenByPasswordError` catches still work identically                                                                                          
  - `cause` gains two new optional fields (`mfa_token`, `mfa_requirements`) that are `undefined` for non-MFA errors                                                      
  - The `isMfaRequiredError` type guard is additive — consumers opt-in to the new behavior                                                                               
                                                                                                                                                                         
  ## How it works                                                                                                                                                        
                                                                                                                                                                         
  Server returns 403:                                                                                                                                                    
  { error: "mfa_required", error_description: "...", mfa_token: "Fe26...", mfa_requirements: {...} }                                                                   
                                                                                                                                                                         
  openid-client throws ResponseBodyError with full response body in .cause                                                                                               
                                                                                                                                                                         
  auth-client.ts catch block calls toOAuth2Error(e):                                                                                                                     
    → detects mfa_required, extracts mfa_token and mfa_requirements from response body                                                                                 
    → returns OAuth2Error with all fields populated                                                                                                                      
                                                                                                                                                                         
  Existing error class (e.g. TokenByPasswordError) is thrown with enriched cause                                                                                         
                                                                                                                                                                         
  Consumer catches error and uses isMfaRequiredError(error) to detect + narrow                                                                                           
                                                                                                                                                                       
  ## Examples                                                                                                                                                            
                                                                       
  ### Basic: detect MFA and start challenge                                                                                                                              
                                                                       
  ```typescript                                                                                                                                                          
  import { AuthClient, isMfaRequiredError } from '@auth0/auth0-auth-js';
                                                                                                                                                                         
  const authClient = new AuthClient({ domain, clientId, clientSecret });                                                                                                 
                                                                                                                                                                         
  try {                                                                                                                                                                  
    await authClient.getTokenByPassword({ username: 'user@example.com', password: 'p4ssw0rd' });                                                                       
  } catch (error) {                                                                                                                                                      
    if (isMfaRequiredError(error)) {                                                                                                                                     
      // error.cause.mfa_token is narrowed to `string`                                                                                                                   
      const challenge = await authClient.mfa.challengeAuthenticator({                                                                                                    
        mfaToken: error.cause.mfa_token,                                                                                                                                 
        challengeType: 'otp',                                                                                                                                            
      });                                                                                                                                                                
    }                                                                                                                                                                    
  }                                                                                                                                                                      
                                                                                                                                                                       
  Refresh token step-up MFA                                                                                                                                              
                                                                       
  import { isMfaRequiredError } from '@auth0/auth0-auth-js';                                                                                                             
                                                                                                                                                                         
  try {                                                                                                                                                                
    await authClient.getTokenByRefreshToken({ refreshToken: 'rt_...' });                                                                                                 
  } catch (error) {                                                                                                                                                      
    if (isMfaRequiredError(error)) {                                                                                                                                     
      const challenge = await authClient.mfa.challengeAuthenticator({                                                                                                    
        mfaToken: error.cause.mfa_token,                                                                                                                                 
        challengeType: 'otp',                                                                                                                                            
      });                                                                                                                                                                
    }                                                                                                                                                                    
  }                                                                                                                                                                    
                                                                                                                                                                         
  Enrollment vs. challenge branching                                                                                                                                     
                                                                                                                                                                       
  import { isMfaRequiredError } from '@auth0/auth0-auth-js';                                                                                                             
                                                                                                                                                                         
  try {                                                                                                                                                                
    await authClient.getTokenByPassword({ username, password });                                                                                                         
  } catch (error) {                                                                                                                                                      
    if (isMfaRequiredError(error)) {                                                                                                                                     
      const { mfa_token, mfa_requirements } = error.cause;                                                                                                               
                                                                                                                                                                         
      if (mfa_requirements?.enroll?.length) {                                                                                                                            
        // User must enroll a new factor                                                                                                                                 
        const enrollment = await authClient.mfa.enrollAuthenticator({                                                                                                    
          mfaToken: mfa_token,                                                                                                                                           
          authenticatorTypes: ['otp'],                                                                                                                                   
        });                                                                                                                                                              
        // Show QR code via enrollment.barcodeUri                                                                                                                        
      } else if (mfa_requirements?.challenge?.length) {                                                                                                                  
        // User has factors enrolled, initiate challenge                                                                                                                 
        await authClient.mfa.challengeAuthenticator({                                                                                                                    
          mfaToken: mfa_token,                                                                                                                                           
          challengeType: mfa_requirements.challenge[0].type as 'otp' | 'oob',                                                                                            
        });                                                                                                                                                              
      }                                                                                                                                                                  
    }                                                                                                                                                                    
  }                                                                                                                                                                    
                                                                                                                                                                         
  Existing error handling still works (no breaking change)                                                                                                               
                                                                                                                                                                       
  import { TokenByPasswordError } from '@auth0/auth0-auth-js';                                                                                                           
                                                                                                                                                                         
  try {                                                                                                                                                                
    await authClient.getTokenByPassword({ username, password });                                                                                                         
  } catch (error) {                                                                                                                                                      
    if (error instanceof TokenByPasswordError) {                                                                                                                         
      // This still catches mfa_required errors — nothing changes for existing consumers                                                                                 
      console.log(error.cause?.error); // 'mfa_required' | 'invalid_grant' | ...                                                                                         
    }                                                                                                                                                                    
  }  
```

### DIFF
--- a/packages/auth0-auth-js/src/auth-client.ts
+++ b/packages/auth0-auth-js/src/auth-client.ts
@@ -6,7 +6,6 @@ import {
   BuildAuthorizationUrlError,
   BuildLinkUserUrlError,
   BuildUnlinkUserUrlError,
-  MfaRequiredError,
   TokenExchangeError,
   MissingClientAuthError,
   NotSupportedError,
@@ -18,7 +17,6 @@ import {
   TokenByRefreshTokenError,
   TokenForConnectionError,
   VerifyLogoutTokenError,
-  type MfaRequirements,
 } from './errors.js';
 import { stripUndefinedProperties } from './utils.js';
 import { MfaClient } from './mfa/mfa-client.js';
@@ -142,17 +140,18 @@ function validateSubjectToken(token: string): void {
   }
 }
 
-function throwIfMfaRequired(e: unknown): void {
+function toOAuth2Error(e: unknown): OAuth2Error {
   const err = e as { error?: string; error_description?: string; cause?: Record<string, unknown> };
-  if (err.error === 'mfa_required') {
-    const cause: OAuth2Error = {
-      error: err.error,
-      error_description: err.error_description ?? 'Multifactor authentication required',
-    };
-    const mfaToken = err.cause?.mfa_token as string | undefined;
-    const mfaRequirements = err.cause?.mfa_requirements as MfaRequirements | undefined;
-    throw new MfaRequiredError('Multifactor authentication required', cause, mfaToken, mfaRequirements);
+  const base: OAuth2Error = {
+    error: err.error ?? '',
+    error_description: err.error_description ?? '',
+    message: (e as { message?: string }).message,
+  };
+  if (err.error === 'mfa_required' && err.cause) {
+    base.mfa_token = err.cause.mfa_token as string | undefined;
+    base.mfa_requirements = err.cause.mfa_requirements as OAuth2Error['mfa_requirements'];
   }
+  return base;
 }
 
 /**
@@ -710,10 +709,9 @@ export class AuthClient {
 
       return TokenResponse.fromTokenEndpointResponse(tokenEndpointResponse);
     } catch (e) {
-      throwIfMfaRequired(e);
       throw new TokenExchangeError(
         `Failed to exchange token for connection '${options.connection}'.`,
-        e as OAuth2Error
+        toOAuth2Error(e)
       );
     }
   }
@@ -768,10 +766,9 @@ export class AuthClient {
 
       return TokenResponse.fromTokenEndpointResponse(tokenEndpointResponse);
     } catch (e) {
-      throwIfMfaRequired(e);
       throw new TokenExchangeError(
         `Failed to exchange token of type '${options.subjectTokenType}'${options.audience ? ` for audience '${options.audience}'` : ''}.`,
-        e as OAuth2Error
+        toOAuth2Error(e)
       );
     }
   }
@@ -915,10 +912,9 @@ export class AuthClient {
 
       return TokenResponse.fromTokenEndpointResponse(tokenEndpointResponse);
     } catch (e) {
-      throwIfMfaRequired(e);
       throw new TokenByRefreshTokenError(
         'The access token has expired and there was an error while trying to refresh it.',
-        e as OAuth2Error
+        toOAuth2Error(e)
       );
     }
   }
@@ -986,10 +982,9 @@ export class AuthClient {
 
       return TokenResponse.fromTokenEndpointResponse(tokenEndpointResponse);
     } catch (e) {
-      throwIfMfaRequired(e);
       throw new TokenByPasswordError(
         'There was an error while trying to request a token.',
-        e as OAuth2Error
+        toOAuth2Error(e)
       );
     }
   }
--- a/packages/auth0-auth-js/src/auth-client.ts
+++ b/packages/auth0-auth-js/src/auth-client.ts
@@ -6,6 +6,7 @@ import {
   BuildAuthorizationUrlError,
   BuildLinkUserUrlError,
   BuildUnlinkUserUrlError,
+  MfaRequiredError,
   TokenExchangeError,
   MissingClientAuthError,
   NotSupportedError,
@@ -17,6 +18,7 @@ import {
   TokenByRefreshTokenError,
   TokenForConnectionError,
   VerifyLogoutTokenError,
+  type MfaRequirements,
 } from './errors.js';
 import { stripUndefinedProperties } from './utils.js';
 import { MfaClient } from './mfa/mfa-client.js';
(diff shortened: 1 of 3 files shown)

### CODE REVIEW

File: packages/auth0-auth-js/src/auth-client.ts [hunk 1]
```diff
@@ -140,6 +140,20 @@ function validateSubjectToken(token: string): void {
...
 }
 
+function toOAuth2Error(e: unknown): OAuth2Error {
+  const err = e as { error?: string; error_description?: string; cause?: Record<string, unknown> };
+  const base: OAuth2Error = {
+    error: err.error ?? '',
+    error_description: err.error_description ?? '',
+    message: (e as { message?: string }).message,
+  };
+  if (err.error === 'mfa_required' && err.cause) {
```
nandan-bhat, line 150: Here we are just checking if `err.cause` exists or not.

Should this function validate that err.cause is actually an object (not just truthy) before accessing properties on it ?

What happens if cause is a string or number, wouldn't `err.cause.mfa_token` throw or return unexpected values ?

Since `e` is `unknown` from the catch block, the as cast on line `143` doesn't provide runtime safety.

Reference: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Error/cause#description

packages/auth0-auth-js/src/auth-client.ts
```diff
@@ -140,6 +140,20 @@ function validateSubjectToken(token: string): void {
...
+    base.mfa_token = err.cause.mfa_token as string | undefined;
```
nandan-bhat, line 151: Nit:
Can we add a typeof check here instead of casting ? Something like `typeof err.cause.mfa_token === 'string' ? err.cause.mfa_token : undefined` ?

This way we don't silently assign a non-string value if the server response shape changes.

packages/auth0-auth-js/src/errors.ts
```diff
@@ -1,10 +1,21 @@
+/**
+ * Describes which MFA factors the user must challenge or enroll.
+ */
+export interface MfaRequirements {
+  challenge?: Array<{ type: string }>;
```
nandan-bhat, line 5: Should type be a string literal union (e.g. `otp' | 'oob' | 'webauthn-roaming' | 'webauthn-platform' | 'push-notification' | 'phone' | 'email'`) instead of plain string ?

That way consumers get autocomplete and don't need to cast when branching on the type.

packages/auth0-auth-js/src/errors.ts
```diff
@@ -173,6 +186,49 @@ export class BuildUnlinkUserUrlError extends ApiError {
...
+ * }
+ * ```
+ */
+export function isMfaRequiredError(
+  error: unknown
+): error is { cause: OAuth2Error & { error: 'mfa_required'; mfa_token: string } } & Error {
+  return (
+    error instanceof Error &&
+    (error as { cause?: OAuth2Error }).cause?.error === 'mfa_required' &&
+    typeof (error as { cause?: OAuth2Error }).cause?.mfa_token === 'string'
```
nandan-bhat, line 228: The return type says `{ cause: OAuth2Error & { error: 'mfa_required'; mfa_token: string } } & Error` but should this also include `mfa_requirements` as optional ?

Without it, consumers who destructure `error.cause` won't see `mfa_requirements` in the narrowed type unless they separately cast it.

packages/auth0-auth-js/src/auth-client.ts
nandan-bhat, line 881: `getTokenByCode` still uses e as `OAuth2Error` instead of toOAuth2Error(e). Is there a reason this method is excluded ? If a user hits step-up MFA after a code exchange, the MFA context on the error would be lost. Even if it's not expected today, using `toOAuth2Error` here would keep all catch blocks consistent.

nandan-bhat, line 1016: Same as 
- https://github.com/auth0/auth0-auth-js/pull/169/changes#r3257102836

packages/auth0-auth-js/src/errors.ts
```diff
@@ -173,6 +186,49 @@ export class BuildUnlinkUserUrlError extends ApiError {
...
+ *       mfaToken: error.cause.mfa_token,
+ *       challengeType: 'otp',
+ *     });
+ *   }
+ * }
+ * ```
+ */
+export function isMfaRequiredError(
+  error: unknown
+): error is { cause: OAuth2Error & { error: 'mfa_required'; mfa_token: string } } & Error {
```
nandan-bhat, line 224: The return type narrows to `{ cause: OAuth2Error & { ... } } & Error`. But `ApiError` declares cause as public `cause?: OAuth2Error` (optional).

After the guard passes, `TypeScript` will believe cause is definitely defined but the narrowed type is `& Error`, `not & ApiError`. This means consumers lose access to `.code`, `.name`, and other `ApiError` properties after narrowing.

Was this intentional ? Would narrowing to `& ApiError` (or a union of the known error classes) be better for DX ?

nandan-bhat:
We may need to add coverage for:

- `toOAuth2Error` with an `mfa_required` error (does it extract the fields correctly?)
- `toOAuth2Error` with a non-MFA error (does it leave `mfa_token` undefined?)
- `isMfaRequiredError` returning true for a valid MFA error
- `isMfaRequiredError` returning false for a regular error, or when `mfa_token` is missing
End-to-end: calling getTokenByPassword when the server returns `403 mfa_required`, and confirming the caught error passes the type guard

subhankarmaiti:
closing this as we have all change sin a single PR https://github.com/auth0/auth0-auth-js/pull/164

PR #170, opened 2026-05-18
fix(auth0-api-js): extract tokens with DPoP scheme using getToken

nandan-bhat:
`getToken` now accepts `Authorization: DPoP <token>` in addition to `Authorization: Bearer <token>`, per `RFC 9449`
Previously, requests using the `DPoP` authorization scheme would throw `InvalidRequestError` since the regex only matched `Bearer`

### Changes
- Broadened the Authorization header regex from `/^Bearer (.+)$/i to /^(?:Bearer|DPoP) (.+)$/i`
- Updated error message from `"No Bearer token found"` to `"No access token found"` (scheme-neutral)
- Added explicit return type annotation to `getTokenFromHeader`
- Added test coverage for `DPoP` scheme (case variations, rejection cases, multiple-method validation)

### DIFF
--- a/packages/auth0-api-js/src/token.spec.ts
+++ b/packages/auth0-api-js/src/token.spec.ts
@@ -17,17 +17,47 @@ describe('getToken', () => {
     });
 
     test.each([
+      { case: 'standard DPoP', authHeader: `DPoP ${validToken}` },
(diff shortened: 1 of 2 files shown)


Piyush-85 (approved):
LGTM

PR #175, opened 2026-05-25
feat(auth0-auth-js): add actor token support and act claim to Custom Token Exchange

cschetan77:
### Description

Extends the Custom Token Exchange (CTE) implementation in `auth0-auth-js` to support RFC 8693 delegation flows via actor tokens.

- Adds `actorToken` and `actorTokenType` optional fields to `ExchangeProfileOptions`, forwarded as `actor_token` and `actor_token_type` in the token request. Both remain in `PARAM_DENYLIST` to prevent injection via `extra` — callers must use the typed fields.
- Adds the `ActClaim` interface `({ sub: string; [key: string]: unknown })` per RFC 8693 Section 4.1. The index signature allows for additional server-provided fields (e.g. chained delegation, `iss`).
- Adds `act?: ActClaim` to `TokenResponse`, populated exclusively inside `#exchangeProfileToken()` (not in the shared `fromTokenEndpointResponse()`). Prefers the `act` claim from the verified ID token; falls back to decoding the JWT access token for M2M flows where no ID token is issued. Wrapped in try/catch to handle opaque access tokens gracefully.
- No client-side URI validation for `subjectTokenType` or `actorTokenType` — invalid URIs are rejected by the Auth0 server, consistent with Swift, Android, and spa-js behaviour.

### Examples

Added a full "Custom Token Exchange" section to `EXAMPLES.md` covering basic exchange, delegation with actor token, reading `act`, M2M fallback behaviour, and error handling.

#### Unit Tests

- should send actor_token and actor_token_type when provided: verifies request params and `result.act` are correctly wired
- should expose act claim from id_token on TokenResponse: verifies extra fields (e.g. `iss`) on `act` are accessible
- should not send actor_token or actor_token_type when omitted: verifies no regression on existing exchanges
- should throw TokenExchangeError when actorToken is provided without actorTokenType
- should not set act when access token is opaque and no id_token is returned
- should prefer act from id_token over act from access token when both are present
- should expose act claim from access token when no id_token is returned (M2M delegation)

#### Manual Tests

- Tested end-to-end against a live Auth0 tenant using a locally built version.
- The test covered a delegation exchange where a service actor exchanges a user's token on their behalf using `actorToken` and `actorTokenType` parameters. The following behaviours were verified:

  - The exchange completes successfully and returns a valid access token and ID token
  - The `act` claim is present on the `TokenResponse`, correctly populated from the ID token with the actor identity set by the CTE Action
  - The refresh token is suppressed by Auth0 when an `actor_token` is present in the request, and the SDK handles the absence gracefully

### Checklist

- [x] All active GitHub checks for tests, formatting, and security are passing

### DIFF
--- a/packages/auth0-auth-js/src/auth-client.ts
+++ b/packages/auth0-auth-js/src/auth-client.ts
@@ -853,9 +853,6 @@ export class AuthClient {
           tokenResponse.act = tokenResponse.claims.act as ActClaim;
         } else {
           try {
-            // The access token is not verified here — the client is not its audience and cannot
-            // validate the signature. The token was received directly from Auth0 over TLS, so
-            // the act claim is trusted in the same way as any other token endpoint response field.
             tokenResponse.act = decodeJwt(tokenEndpointResponse.access_token).act as ActClaim | undefined;
           } catch {
             // opaque access token — act claim not available
--- a/packages/auth0-auth-js/src/auth-client.ts
+++ b/packages/auth0-auth-js/src/auth-client.ts
@@ -78,6 +78,8 @@ const MAX_ARRAY_VALUES_PER_KEY = 20;
  *   credentials must be managed through configuration, not request parameters
  * - subject_token, subject_token_type: Core token exchange parameters, overriding creates
  *   ambiguity about which token is being exchanged
+ * - actor_token, actor_token_type: Actor token parameters for delegation exchanges, must use
+ *   explicit typed parameters to ensure correct delegation semantics
  * - requested_token_type: Determines the type of token returned, must be explicit
  * - audience, aud, resource, resources, resource_indicator: Target API parameters must use
  *   explicit API parameters to prevent confusion about precedence and ensure correct routing
@@ -101,6 +103,8 @@ const PARAM_DENYLIST = Object.freeze(
     'subject_token',
     'subject_token_type',
     'requested_token_type',
+    'actor_token',
+    'actor_token_type',
     'audience',
     'aud',
     'resource',
@@ -157,18 +161,6 @@ function toOAuth2Error(e: unknown): OAuth2Error {
     }
   }
   return base;
-}
-
-/**
- * Validates that a token type value is a syntactically valid URI.
- * Semantic validation (reserved namespaces) is enforced by the Auth0 server.
- */
-function validateTokenTypeUri(value: string, paramName: string): void {
-  try {
-    new URL(value);
-  } catch {
-    throw new TokenExchangeError(`${paramName} must be a valid URI`);
-  }
 }
 
 /**
@@ -817,10 +809,9 @@ export class AuthClient {
     const { configuration } = await this.#discover();
 
     validateSubjectToken(options.subjectToken);
-    validateTokenTypeUri(options.subjectTokenType, 'subjectTokenType');
 
-    if (options.actorTokenType !== undefined) {
-      validateTokenTypeUri(options.actorTokenType, 'actorTokenType');
+    if (options.actorToken !== undefined && options.actorTokenType === undefined) {
+      throw new TokenExchangeError('actorTokenType is required when actorToken is provided');
     }
 
     const tokenRequestParams = new URLSearchParams({
--- a/packages/auth0-auth-js/src/auth-client.ts
+++ b/packages/auth0-auth-js/src/auth-client.ts
@@ -78,7 +78,6 @@ const MAX_ARRAY_VALUES_PER_KEY = 20;
  * - subject_token, subject_token_type: Core token exchange parameters, overriding creates
  *   ambiguity about which token is being exchanged
  * - requested_token_type: Determines the type of token returned, must be explicit
- * - actor_token, actor_token_type: Delegation parameters that affect authorization context
  * - audience, aud, resource, resources, resource_indicator: Target API parameters must use
  *   explicit API parameters to prevent confusion about precedence and ensure correct routing
  * - scope: Overriding via extras bypasses the explicit scope parameter and creates ambiguity
@@ -101,8 +100,6 @@ const PARAM_DENYLIST = Object.freeze(
     'subject_token',
     'subject_token_type',
     'requested_token_type',
-    'actor_token',
-    'actor_token_type',
     'audience',
     'aud',
     'resource',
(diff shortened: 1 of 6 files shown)


packages/auth0-auth-js/src/types.ts
```diff
@@ -632,6 +686,8 @@ export class TokenResponse {
 
     tokenResponse.tokenType = response.token_type;
     tokenResponse.issuedTokenType = (response as typeof response & { issued_token_type?: string }).issued_token_type;
+    const atClaims = decodeJwt(response.access_token);
```
nandan-bhat: What will happen if Auth0 issues an opaque token.

packages/auth0-auth-js/src/types.ts
```diff
@@ -632,6 +686,8 @@ export class TokenResponse {
...
+    tokenResponse.act = (claims?.act ?? atClaims.act) as ActClaim | undefined;
```
nandan-bhat: Does that mean, `tokenResponse` would always have `act` ?
Because `fromTokenEndpointResponse` is called from multiple places.

nandan-bhat:
- Please add tests for opaque token scenarios.
- Please update documentation (EXAMPLES.md) elaborating this feature.

packages/auth0-auth-js/src/types.ts
```diff
@@ -534,6 +561,23 @@ export interface AuthorizationDetails {
...
 
+/**
+ * Represents the `act` (actor) claim in a token response (RFC 8693).
+ *
+ * Present when a token was issued via a delegation exchange, identifying the
+ * acting party (e.g., an intermediate service) that performed the exchange on
+ * behalf of the subject.
+ *
+ * @see {@link https://www.rfc-editor.org/rfc/rfc8693#section-4.1 RFC 8693 Section 4.1}
+ */
```
kishore7snehil, line 571: The `act` claim is defined in **Section 4.4** ("Delegation Semantics"), not 4.1. Section 4.1 is the token exchange response structure.

Should be `#section-4.4`. (Same fix landed on the Android and Swift CTE PRs.)

packages/auth0-auth-js/src/types.ts
```diff
@@ -589,6 +633,16 @@ export class TokenResponse {
...
   recoveryCode?: string;
 
+  /**
+   * The actor claim from the ID token (RFC 8693).
+   *
+   * Present in delegation exchanges when an `actorToken` was provided. Identifies
+   * the acting party on whose behalf the subject token was exchanged.
+   *
+   * @see {@link https://www.rfc-editor.org/rfc/rfc8693#section-4.1 RFC 8693 Section 4.1}
+   */
```
kishore7snehil, line 644: Same as the `ActClaim` interface above: `act` is defined in **Section 4.4**, not 4.1. Should be `#section-4.4`.

packages/auth0-auth-js/src/auth-client.ts
```diff
@@ -757,6 +766,14 @@ export class AuthClient {
     const { configuration } = await this.#discover();
 
     validateSubjectToken(options.subjectToken);
+    validateTokenTypeUri(options.subjectTokenType, 'subjectTokenType');
+
```
kishore7snehil, line 812: Two concerns here:

- The team aligned that token-type URI validation stays server-side,  no other SDK (Swift, Android, spa-js) does client-side URI validation. Adding it here for `subjectTokenType` (which previously had none) breaks that alignment and is a behavior change for existing callers.

- **`new URL()` is the wrong validator and risks false rejections.** It can reject values the Auth0 server would actually accept, turning a server-side concern into a client-side breaking change on upgrade. (`new URL('urn:acme:custom-token')` happens to pass, but the general class of valid token-type identifiers isn't guaranteed to parse as a WHATWG URL.)

packages/auth0-auth-js/src/auth-client.ts
```diff
@@ -101,8 +100,6 @@ const PARAM_DENYLIST = Object.freeze(
     'subject_token',
     'subject_token_type',
     'requested_token_type',
-    'actor_token',
-    'actor_token_type',
```
kishore7snehil, line 105: A caller can again pass `actor_token` through `extra`, which is no longer guarded. The typed param is the documented path.

kishore7snehil (changes requested):

packages/auth0-auth-js/src/types.ts
cschetan77: yes, this condition needs to be handled. Updated to catch the error when it's a opaque token and populate the act claim to be undefined.

packages/auth0-auth-js/src/types.ts
cschetan77: yes, tokenResponse from any exchange would populate the `act` claim, but it will be `undefined` except response from `exchangeProfileToken`. 

I agree, the static method `fromTokenEndpointResponse` is called at many places. And none of them need the `act` claim population logic except `exchangeProfileToken`. 
It seems, a better design choice will be to keep the `act` claim population logic in that specific method itself. 

**Update:** The `act` claim population logic has been - 
1. Removed  from `fromTokenEndpointResponse` static method. 
2. Added in `exchangeProfileToken` method, which is the only method that would ever populate it.

What do you suggest!

packages/auth0-auth-js/src/types.ts
cschetan77: I double checked the RFC 8693, the act claim appears to be defined in `Section 4.1`, titled "act" (Actor) Claim. 
Section 4.4 ("Delegation Semantics") references the claim as part of the broader delegation flow description, but the formal definition  seems to live in 4.1.
Though maybe I'm missing something here or not seeing the full context.

packages/auth0-auth-js/src/auth-client.ts
cschetan77: Yes, that makes complete sense. 
Removed the entire URI validation logic.

packages/auth0-auth-js/src/auth-client.ts
cschetan77: yes, these shouldn't have been removed. 
I've added them back to the denylist so the typed params remain the only documented and enforced path.

kishore7snehil (dismissed):
LGTM!

nandan-bhat (approved):

PR #176, opened 2026-05-25
feat(auth0-server-js): add loginWithCustomTokenExchange and customTokenExchange methods

cschetan77:
### Description

Adds first-class Custom Token Exchange (CTE) support to ServerClient in auth0-server-js, building on the actor token and act claim foundation added to auth0-auth-js.

- `loginWithCustomTokenExchange(options, storeOptions?)`: exchanges a custom external token (e.g. a Google ID token, a legacy system token) for Auth0 tokens and persists the result as a user session in the StateStore. Equivalent to completing an interactive login flow, but without a browser redirect. Calls stateStore.set with removeIfExists: true to prevent session fixation. Returns `Promise<LoginWithCustomTokenExchangeResult>`, surfacing authorizationDetails from the token response for callers using RAR (authorization_details in extra). 
- `customTokenExchange(options, storeOptions?)`: performs the same RFC 8693 token exchange but returns the raw TokenResponse without touching the StateStore. Designed for delegation and impersonation flows where downstream tokens are needed but no session should be created or modified.
- `LoginWithCustomTokenExchangeOptions` and `CustomTokenExchangeOptions` type aliases in types.ts, both derived from `ExchangeProfileOptions` in `auth0-auth-js`. This keeps the types consistent across the monorepo and inherits `actorToken`, `actorTokenType`, and all other exchange options automatically. Also added `LoginWithCustomTokenExchangeResult` interface to types.ts

Both methods follow the same domain resolution and AuthClient routing pattern used by loginBackchannel, supporting both static domain and resolver mode configurations. 
  
The return of `Promise<LoginWithCustomTokenExchangeResult>` containing `authorizationDetails` is kept in consistent with `completeInteractiveLogin` and `loginBackchannel` methods. Though the case for returning it is narrow.

#### Unit tests 
 - loginWithCustomTokenExchange - should persist session after successful exchange
 - loginWithCustomTokenExchange - should use "default" audience when none provided
 - loginWithCustomTokenExchange - should persist domain on session
 - loginWithCustomTokenExchange - should call stateStore.set with removeIfExists=true (session fixation)
 - loginWithCustomTokenExchange - should throw when exchange fails
 - loginWithCustomTokenExchange - should return authorizationDetails when RAR was used
 - loginWithCustomTokenExchange - should resolve domain via resolver function
 - loginWithCustomTokenExchange - should allow getAccessToken to return the token after login
 - loginWithCustomTokenExchange - should persist act claim on session user when actor token is used
 - customTokenExchange - should return token response without persisting session
 - customTokenExchange - should throw when exchange fails
 - customTokenExchange - should return act claim when actor token is used

##### customTokenExchange(): stateless delegation flow
- Called with a subject token and an actor token
- Exchange succeeded and returned a TokenResponse
- `act.sub` populated correctly (set by api.authentication.setActor() in the Action)
- refreshToken correctly absent, suppressed when actor_token is present per RFC 8693

##### loginWithCustomTokenExchange(): session-persisting
- Called with a subject token only (no actor token)
- Exchange succeeded and session was written to the state store
- session.user confirmed to be the subject token user
- accessToken, idToken, and refreshToken all present in the session

### Checklist

- [ ] The correct base branch is being used, if not the default branch

## Release Notes

* **New Features**
  * Added custom token exchange APIs (RFC 8693) for exchanging external tokens into Auth0 sessions.
  * Supports actor-based delegation, and exposes returned authorization details when using RAR.
  * Added a non-session “raw” exchange mode that returns token data without persisting session state.

* **Documentation**
  * Added a “Login using Custom Token Exchange” guide, including required scope behavior and `storeOptions` usage.

* **Tests**
  * Added coverage for session persistence, domain resolution via resolvers, and error handling for failed exchanges.


packages/auth0-server-js/src/server-client.ts
```diff
@@ -705,6 +709,73 @@ export class ServerClient<TStoreOptions = unknown> {
...
+    options: LoginWithCustomTokenExchangeOptions,
+    storeOptions?: TStoreOptions
+  ): Promise<LoginWithCustomTokenExchangeResult> {
+    const domain = await this.#resolveDomain(storeOptions);
+    const authClient = this.#getAuthClient(domain);
+    const tokenEndpointResponse = await authClient.exchangeToken(options);
+
+    const existingStateData = await this.#stateStore.get(this.#stateStoreIdentifier, storeOptions);
+    const stateData = updateStateData(
+      options.audience ?? 'default',
```
nandan-bhat: Here we save the token set using the audience passed in the call `(options.audience)`. But `getAccessToken` reads the token set using the audience from the client config `(this.#options.authorizationParams?.audience)`. `loginBackchannel` and `completeInteractiveLogin` both save using the config audience, not a per call value.

So if someone calls this method with audience `'https://api.example.com'` but the client is not configured with that same `audience`, `getAccessToken` will look under `'default'`, not find the saved token, and try to refresh. When an actor token is used there is no refresh token, so it will throw.

The `EXAMPLES` doc tells people to call `getAccessToken` after this, so this path can break. Can we either save under the config audience like the other login methods, or make the contract clear ? It would also help to add a test that logs in and then calls getAccessToken to prove it works end to end.

packages/auth0-server-js/src/server-client.ts
```diff
@@ -705,6 +709,73 @@ export class ServerClient<TStoreOptions = unknown> {
...
+   *
+   * @returns A promise resolving to an object containing `authorizationDetails` when RAR was used.
+   */
+  public async loginWithCustomTokenExchange(
+    options: LoginWithCustomTokenExchangeOptions,
+    storeOptions?: TStoreOptions
+  ): Promise<LoginWithCustomTokenExchangeResult> {
+    const domain = await this.#resolveDomain(storeOptions);
+    const authClient = this.#getAuthClient(domain);
+    const tokenEndpointResponse = await authClient.exchangeToken(options);
```
nandan-bhat: The other login methods (`loginBackchannel`, `startInteractiveLogin`) call `ensureOpenIdScope` before the exchange. This one passes options straight through.

This is a login method and the session user is taken from the id token claims. The user is only set when an id token comes back. If the caller forgets `openid` in the `scope`, `Auth0` may not return an id token, and we end up with a logged in session where user is undefined. The tests do not catch this because the mock always returns an id token.

Can we either add openid scope here like the sibling methods, or document that `openid` is required ?  `customTokenExchange` can stay as is since it does not create a session.

packages/auth0-server-js/src/server-client.spec.ts
```diff
@@ -2228,6 +2228,244 @@ test('loginBackchannel - should not duplicate openid when already present in cus
   spy.mockRestore();
 });
 
+test('loginWithCustomTokenExchange - should persist session after successful exchange', async () => {
```
nandan-bhat, line 2231: None of the new tests pass `actorToken` or `actorTokenType`, and none check the act claim. Actor tokens are the main point of this feature.

The spec says the act claim must be saved in the session for `loginWithCustomTokenExchange` and must be readable on the response for `customTokenExchange`. The actor tests today only exist at the auth-client level, not for these new server client methods.

Can we add two tests:
1. `loginWithCustomTokenExchange` with an actor token, then check the saved session user has act.
2. `customTokenExchange` with an actor token, then check `result.act.sub`.

packages/auth0-server-js/EXAMPLES.md
```diff
@@ -904,6 +908,72 @@ await serverClient.loginBackchannel({}, storeOptions);
...
+  subjectTokenType: 'urn:ietf:params:oauth:token-type:access_token',
+  audience: 'https://downstream-api.example.com',
+});
+
+// Use tokenResponse.accessToken to call the downstream API
+```
+
+No StateStore reads or writes occur — `storeOptions` is only used for domain resolution in resolver mode.
+
+### Using actor tokens for delegation
```
nandan-bhat, line 947: The spec says we should document that no refresh token is given when an actor token is sent, even if `offline_access` is in the `scope`. This section does not mention it.

Without this note, someone using an actor token with `loginWithCustomTokenExchange` will call `getAccessToken` after the token expires and hit an error with no warning. Can we add a short line here that says refresh tokens are not issued for actor token exchanges, so the session cannot be refreshed silently ?

packages/auth0-server-js/EXAMPLES.md
```diff
@@ -37,21 +37,25 @@
 - [Login using Client-Initiated Backchannel Authentication](#login-using-client-initiated-backchannel-authentication)
   - [Using Rich Authorization Requests](#using-rich-authorization-requests)
   - [Passing `StoreOptions`](#passing-storeoptions-2)
-- [Retrieving the logged-in User](#retrieving-the-logged-in-user)
+- [Login using Custom Token Exchange](#login-using-custom-token-exchange)
+  - [Performing a delegation exchange without a session](#performing-a-delegation-exchange-without-a-session)
+  - [Using actor tokens for delegation](#using-actor-tokens-for-delegation)
   - [Passing `StoreOptions`](#passing-storeoptions-3)
```
nandan-bhat: These `StoreOptions` anchors do not match the slugs `GitHub` builds. `GitHub` numbers repeated headings by their order in the page. This new section is the 6th `'Passing StoreOptions'` heading, so the link should be `passing-storeoptions-5`, not -3. The backchannel one above also looks off.

This numbering was already messy before this PR, but the new section keeps the pattern. Can we fix the anchors so the links actually jump to the right place ?

nandan-bhat (commented):
Reviewed the custom token exchange methods. The change is clean and the tests pass. I left a few inline notes. The main one is the audience used when we save the token set, since it may not match what getAccessToken looks for later. I also think we should add tests that pass an actor token and check the act claim, since that is the main goal of this feature.

packages/auth0-server-js/src/server-client.ts
cschetan77: yes, I agree, this would create a subtle bug, a session would be silently created without any error, with id token claims undefined. 
Update: Added
- `ensureOpenIdScope` check in `loginWithCustomTokenExchange`
- a note in `EXAMPLES.md` to clearly call out that SDK requires `openid` scope for this method, if not provided, SDK will inject appropriate defaults.

packages/auth0-server-js/src/server-client.ts
cschetan77: yes, we should key the accessToken in TokenSets under primary config audience `this.#options.authorizationParams?.audience` or the `default` as we do in `loginBackchannel` and `completeInteractiveLogin`. 
Update:
- The access token gets saved under `this.#options.authorizationParams?.audience ?? 'default'`
- unit test: `loginWithCustomTokenExchange - should allow getAccessToken to return the token after login`- end-to-end test that proves loginWithCustomTokenExchange followed by getAccessToken works.

packages/auth0-server-js/EXAMPLES.md
cschetan77: thanks for pointing this out. 
Fixed the TOC.

packages/auth0-server-js/EXAMPLES.md
cschetan77: Added a important note under this section to clearly call out this.

packages/auth0-server-js/src/server-client.spec.ts
cschetan77: Added 2 proper unit tests 

- loginWithCustomTokenExchange - should persist `act` claim on session user when actor token is used
- customTokenExchange - should return `act` claim when actor token is used

nandan-bhat:
LGTM. Thanks for addressing all the review items.

nandan-bhat (approved):

PR #181, opened 2026-06-08
feat : add passkey signup and login support to auth0-server-js

pmathew92:
### Summary

Adds passkey (WebAuthn) authentication support to `auth0-server-js`, mirroring the approach in `auth0-spa-js`. The actual Auth0 API calls are delegated to `auth0-auth-js` (passkey support shipped there in `1.7.0`); this package wraps those calls and layers in **session management** following this repo's existing `StateStore` model — exactly like `loginBackchannel` does.

Three new methods on `ServerClient`:

| Method | Purpose | Writes session? |
|---|---|---|
| `passkeyRegister(options, storeOptions?)` | Request a signup challenge | No |
| `passkeyChallenge(options?, storeOptions?)` | Request a login challenge | No |
| `passkeyGetToken(options, storeOptions?)` | Exchange the credential for tokens | **Yes** |

Because the WebAuthn ceremony (`navigator.credentials`) can only run in the browser, the flow is a **browser↔server round trip**: `passkeyRegister`/`passkeyChallenge` return a challenge + `authSession` that the app relays to the browser; the browser produces a credential; the app passes it back to `passkeyGetToken`, which exchanges it via the WebAuthn grant and persists the resulting session. The `authSession` round-trips through the browser and is **not** stored server-side.

###  Changes

- **`src/types.ts`** — re-export passkey types from `auth0-auth-js`; add `PasskeyGetTokenResult`.
- **`src/server-client.ts`** — add `passkeyRegister`, `passkeyChallenge`, `passkeyGetToken`. `passkeyGetToken` runs `ensureOpenIdScope()` at token exchange (so an `id_token`/user is always returned) and merges tokens into the session via `updateStateData` → `stateStore.set`.
- **`package.json`** — bump `@auth0/auth0-auth-js` floor `^1.1.0` → `^1.7.0` (first version with passkey support).
- **`src/server-client.spec.ts`** — 5 tests: register/challenge don't write to the state store, `getToken` exchanges + stores tokens, custom scope always includes `openid`, and challenge-endpoint failure propagates.
- **`EXAMPLES.md`** — new "Login and Signup using Passkeys" section documenting the round trip, plus an IMPORTANT callout: passkey auth method + WebAuthn grant must be enabled and a **custom domain is required** (passkeys don't work on `*.auth0.com`).

### Checklist

- [X] I have added documentation for new/changed functionality in this PR or in auth0.com/docs
- [X] All active GitHub checks for tests, formatting, and security are passing
- [X] The correct base branch is being used, if not the default branch

## Release Notes

* **New Features**
  * Added server-side WebAuthn passkey APIs to register, challenge, and complete sign-in via passkeys.
  * Supports token exchange with session/state persistence and returns optional authorization details.

* **Documentation**
  * Updated passkey example navigation and added an end-to-end browser/server passkey workflow, including required configuration guidance and MFA-required handling.

* **Tests**
  * Expanded automated coverage for passkey flows, including error propagation and token exchange behavior.


packages/auth0-server-js/src/types.ts
```diff
@@ -1,6 +1,16 @@
 import type { AuthorizationDetails, DiscoveryCacheOptions, TelemetryConfig } from '@auth0/auth0-auth-js';
 
-export type { DiscoveryCacheOptions } from '@auth0/auth0-auth-js';
+export type {
+  DiscoveryCacheOptions,
```
Piyush-85, line 4: It would be better if we rename types as well to PasskeyRegisterOptions, PasskeyRegisterResponse and others

packages/auth0-server-js/src/server-client.spec.ts
```diff
@@ -4338,3 +4377,130 @@ test('Telemetry - should include Auth0-Client header in token requests', async (
...
+  await serverClient.passkeyGetToken({
+    authSession: 'auth_session_challenge_123',
+    credential: fakePasskeyCredential,
+    scope: 'profile email',
+  });
+
+  expect(capturedScope).toBe('openid profile email');
+});
+
+test('passkeyChallenge - should throw when the challenge endpoint fails', async () => {
```
Piyush-85, line 4817: **Missing test:** `passkeyRegister` failure propagation
`passkeyChallenge` failure is tested but `passkeyRegister` failure is not

Piyush-85:
@pmathew92 Have added a few minor comments, PTAL.
**NOTE:**  Its better tp merge this after https://github.com/auth0/auth0-auth-js/pull/187 is merged as it contains fix for `mfa_required` error handling.

nandan-bhat:
The passkey flow has two server calls with the browser in between. First `passkeyChallenge()` runs on the server and returns an `authSession`. Then the
  browser does the `WebAuthn` ceremony. Then `passkeyGetToken()` runs on the server and needs that same `authSession` back.

  Right now the SDK does not store `authSession` anywhere on the server. So the only way the second call can get it is from the browser. That means the app
  has to send `authSession` down to the browser after the challenge, hold it there during the ceremony, and send it back up for the token exchange.

  This is a bit different from how `startInteractiveLogin` works in this repo. There the SDK saves the `state` and `nonce` in the TransactionStore (a cookie), so the browser never has to carry that value, the SDK just looks it up on the callback.

Because of this difference, each SDK on top of this one (`auth0-express`, `auth0-fastify`, `auth0-nuxt`) will have to write its own code to pass `authSession` to the browser and read it back. If `auth0-server-js` stored it in the `TransactionStore` instead, all of them would get the same behaviour for free.

Is the browser relay the intended design here, or should we store `authSession` in the `TransactionStore` so the flow matches the interactive login pattern  and stays consistent across the framework SDKs ?

nandan-bhat:
The `authSession` is tied to the Auth0 domain that issued it.

In resolver mode the `domain` is a function, so the SDK resolves it separately on each call from the `storeOptions` you pass in. If the resolver returns one domain on the challenge call and a different domain on the `getToken` call, the token exchange will hit a different custom-domain than the one that issued the `authSession`, and it will fail.

A concrete case where this can happen: a multi-domain app that picks the Auth0 domain from the request, for example from the subdomain or a header. Say the challenge call comes in on `custom-domain-a.myapp.com` and resolves to `custom-domain-a.acmetest.com`. The browser does the ceremony. Then the getToken call comes back, but this time it carries a context that resolves to `custom-domain-b.acmetest.com` (wrong subdomain, a stale value, a load balancer rewrite, or the app just builds `storeOptions` differently on the two routes). Now the credential built for custom-domain A is being exchanged against custom-domain B, and it breaks.

For the static domain setup this cannot happen, since the domain is fixed.

packages/auth0-server-js/src/types.ts
pmathew92: Done

packages/auth0-server-js/src/server-client.spec.ts
pmathew92: Added a new test case for `passkeyRegister` failure

Piyush-85 (dismissed):

packages/auth0-server-js/src/server-client.ts
```diff
@@ -499,6 +505,105 @@ export class ServerClient<TStoreOptions = unknown> {
...
+    *
+    * Call this after obtaining a credential from `navigator.credentials.create()`
+    * (signup) or `navigator.credentials.get()` (login), passing the `authSession`
+    * returned by `passkeyRegister()` / `passkeyChallenge()` together with the
+    * serialized credential.
+    *
+    * @param options The auth session, serialized credential, and optional realm/scope/audience/organization.
+    * @param storeOptions Optional options used to pass to the State Store (and to resolve the domain in resolver mode).
+    *
+    * @throws {PasskeyGetTokenError} If there was an issue exchanging the credential for tokens. When the cause is `mfa_required`, use `isMfaRequiredError(error)` to narrow the error and read `cause.mfa_token`. No session is persisted in this case.
```
nandan-bhat: The JSDoc here says these methods throw `PasskeyRegisterError`, `PasskeyChallengeError` and `PasskeyGetTokenError`, but those classes are not exported from `@auth0/auth0-server-js`. They only live in `@auth0/auth0-auth-js`. So if someone wants to do `catch (e) { if (e instanceof PasskeyGetTokenError) ... }` they cannot import the class from this package.

We already re-export the MFA error classes from `mfa/index.ts` for this exact reason, so it feels a bit inconsistent to leave these out. Can we re-export the three passkey error classes the same way ? The `mfa_required` path is fine since `isMfaRequiredError` is already re-exported.

packages/auth0-server-js/src/server-client.spec.ts
```diff
@@ -4667,3 +4706,187 @@ test('Telemetry - should include Auth0-Client header in token requests', async (
...
+  });
+
+  const result = await serverClient.passkeyChallenge();
+
+  expect(result.authSession).toBe('auth_session_challenge_123');
+  expect(result.authnParamsPublicKey.challenge).toBe('login_challenge');
+  expect(mockStateStore.set).not.toHaveBeenCalled();
+});
+
+test('passkeyGetToken - should exchange the credential and store the access token', async () => {
```
nandan-bhat, line 4754: The coverage here is good. It checks scope injection, the store write, the `mfa_required` case, and the challenge/register failures. Two things are not tested though: the `authorizationDetails` return value (which the docs show off for RAR), and forwarding of `organization` / `audience` to the grant.

Could we add a quick test that asserts `authorizationDetails` comes back when the response has `authorization_details`, and one that checks `audience` / `organization` get passed through ? That would lock in the documented behavior.

packages/auth0-server-js/src/server-client.spec.ts
```diff
@@ -65,13 +75,18 @@ const restHandlers = [
...
-          expires_in: 60,
-        });
+        auth_req_id: auth_req_id,
+        interval: 0.5,
+        expires_in: 60,
+      });
   }),
   http.post(mockOpenIdConfiguration.token_endpoint, async ({ request }) => {
-    const info = await request.formData();
+    // The passkey (WebAuthn) grant sends application/json; all other grants
```
nandan-bhat, line 84: Looks like this diff also re-indents the existing `bc-authorize` and `token_endpoint` response objects and adds a couple of empty lines that are not related to passkeys. There is some similar churn over in `server-client.ts` too, with double blank lines between the new methods and a JSDoc block that is indented one space too far.

Can we revert the unrelated formatting changes so the diff only shows the passkey work ? It makes the review easier and keeps the blame clean.

packages/auth0-server-js/src/server-client.ts
pmathew92: Exported the Errortypes

packages/auth0-server-js/src/server-client.spec.ts
pmathew92: Added a new test case for the same

packages/auth0-server-js/src/server-client.spec.ts
pmathew92: Done

nandan-bhat (approved):

PR #190, opened 2026-06-14
feat(auth0-server-js): enforce IPSIE session_expiry as a session ceiling

nandan-bhat:
Enforces the IPSIE `session_expiry` claim in `@auth0/auth0-server-js` as a hard ceiling on the local session. When an `Okta/OIDC` enterprise  connection has `id_token_session_expiry_supported: true`, Auth0 emits a `session_expiry` claim (absolute Unix seconds) in the ID token; this SDK now treats it as an upper bound the session and any refresh renewal must
 never outlive.

**The change is non-breaking** — all enforcement is gated on the claim being present. Sessions without it (database/social logins, the connection option disabled, or sessions created before this version) behave exactly as before.

## What it does
- **Persists the ceiling** as a top-level `SessionData.sessionExpiresAt?:  number` (Unix seconds) at login, so `getSession()` can read it back. Stamped  in `updateStateData` which is the single funnel for interactive login, backchannel login, and MFA-completed login.
- **Lockout guard** at login: if `session_expiry <= iat` (fallback to now), throws `SessionExpiredError` and persists nothing rather than storing a born-dead session.
- **Reads (`getSession` / `getUser`)** — once the ceiling is reached, the current-domain session is deleted and the call returns `undefined` (silent), so the app's existing not-logged-in redirect path fires transparently.
- **Token paths (`getAccessToken` / `getAccessTokenForConnection`)** — once the ceiling is reached, the session is deleted and `SessionExpiredError` is thrown **before** any cached token is served or any `/oauth/token` refresh is attempted (never refresh past the ceiling).
- **`startLinkUser` / `startUnlinkUser`** — same throwing enforcement, since they use the session ID token for account mutation.
- **Refresh carry-forward**: the stored ceiling is preserved across refreshes (`extractSessionExpiry(claims) ?? stored`), so a refresh that doesn't re-assert the claim can't silently remove the limit. capped at the ceiling, so a stale/stolen cookie can't replay past it. Idle/absolute timeouts stay independent.
- **Fail-open**: a missing or malformed claim means "no ceiling," never "already expired" — a platform glitch can't lock out every enterprise user.
- **~30s negative leeway** for clock skew; integer-seconds comparisons throughout.
- Adds `SessionExpiredError` (`code: 'session_expired'`) and a "Session expiry from upstream IdP" guide in EXAMPLES.md + README link.

## Behavior changes for consumers (opt-in, once the feature is enabled)
No API breakage, but two new runtime behaviors surface **only** after a connection enables the option and a user logs in through it:
- `getSession()` / `getUser()` can return `undefined` for a previously-logged-in user once the ceiling passes — apps that assumed these always return a value after login should add a null check.
- `getAccessToken()` / `getAccessTokenForConnection()` / `startLinkUser()` / `startUnlinkUser()` can throw `SessionExpiredError` past the ceiling.

## Public API (all additive)
- New optional field `SessionData.sessionExpiresAt?: number`
- New exported error `SessionExpiredError`
- New optional param on the `protected calculateMaxAge(createdAt,

Nothing removed, renamed, or retyped.

## Test plan
- [x] `extractSessionExpiry` / `isSessionExpiryReached` unit tests (valid/invalid shapes, leeway boundaries)
- [x] `updateStateData`: stamping, lockout, refresh carry-forward, same-user / different-user re-login
- [x] `getSession` / `getUser`: expired → undefined + delete; future/absent → unchanged; resolver-mode foreign domain not deleted
- [x] `getAccessToken` / `getAccessTokenForConnection`: expired → throws `SessionExpiredError`, no `/oauth/token` call, session deleted; future →cached token served (regression)
- [x] `startLinkUser` / `startUnlinkUser`: expired → throws, URL builder never called
- [x] `completeInteractiveLogin` E2E lockout
- [x] Cookie max-age cap (rolling + non-rolling)
- [x] Full suite green, `tsc --noEmit` clean, lint clean, build succeeds

## Summary by CodeRabbit

* **New Features**
  * Sessions now automatically expire based on upstream identity provider claims. Expired sessions are cleared from storage, and attempts to retrieve access tokens or perform account linking operations are denied with appropriate error handling.

* **Documentation**
  * Added comprehensive documentation covering session expiry from upstream identity providers, including configuration prerequisites, behavior in login and token refresh scenarios, and upgrade guidance for existing applications.

### DIFF
--- a/packages/auth0-server-js/EXAMPLES.md
+++ b/packages/auth0-server-js/EXAMPLES.md
@@ -1054,6 +1054,63 @@ const accessToken = await serverClient.getAccessTokenForConnection({}, storeOpti
 
 Read more above in [Configuring the Store](#configuring-the-store)
 
+## Session expiry from upstream IdP (IPSIE `session_expiry`)
+
+When you use an Okta or OIDC **enterprise connection** configured with `id_token_session_expiry_supported: true`, Auth0 includes a `session_expiry` claim in the ID token it issues to your application. This claim is an absolute Unix timestamp (seconds) that marks the latest moment the upstream identity provider considers the session valid. It is computed by Auth0 as the earliest of: your tenant's absolute session lifetime, the upstream IdP's own session expiry, and any value an Action sets via `api.session.setExpiresAt`.
+
+### What the SDK does automatically
+
+No configuration or code change is required. When the claim is present, the SDK:
+
+- persists it on the session as a top-level `sessionExpiresAt` (Unix seconds);
+- treats the session as expired once `sessionExpiresAt` is reached (with a small 30-second leeway for clock skew), on **every** `getSession()` / `getUser()` / `getAccessToken()` / `getAccessTokenForConnection()` call;
+- never refreshes a token once the ceiling has passed; and
+- caps the session cookie lifetime at the ceiling as a defense-in-depth backstop.
+
+This is layered **on top of** your existing idle and absolute session timeouts — it does not replace them. The session ends at whichever limit is reached first.
+
+### Behavior on expiry
+
+- `getSession()` and `getUser()` return `undefined` (the session is also cleared from the store). Your existing "not logged in → redirect to login" path runs unchanged.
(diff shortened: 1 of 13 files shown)


nandan-bhat:
This is a duplicate PR. Closing this.

PR #191, opened 2026-06-14
feat(auth0-server-js): enforce IPSIE session_expiry as a session ceiling

nandan-bhat:
Enforces the IPSIE SL1 `session_expiry` claim in `@auth0/auth0-server-js` as a hard ceiling on the local session. When an Okta/OIDC enterprise connection has `id_token_session_expiry_supported: true`, Auth0 emits a `session_expiry` claim (absolute Unix seconds) in the ID token; this SDK treats it as an upper bound the session — and any refresh renewal — must never outlive.

**The change is non-breaking** — all enforcement is gated on the claim being present. Sessions without it (database/social logins, the connection option disabled, or sessions created before this version) behave exactly as before.

## Architecture

The ceiling is **stamped once, at the login sites** — `completeInteractiveLogin`, `loginBackchannel`, and MFA `verify` — via `applySessionExpiryAtLogin()`, which extracts the claim, applies the born-expired lockout, and stamps a top-level `SessionData.sessionExpiresAt`. `updateStateData` is **preserve-only**: it never derives the ceiling, so it is carried forward unchanged across refreshes via the state spread. This makes it structurally impossible for a refresh-token grant to overwrite the ceiling — even when a Post-Login Action stamps `session_expiry` on the refreshed ID token. (Mirrors `auth0-server-python`.)

## What it does

- **Persists the ceiling** as a top-level `SessionData.sessionExpiresAt?: number` (Unix seconds) at login, so `getSession()` can read it back.
- **Lockout guard** at login: if `session_expiry <= iat` (falling back to now when `iat` is absent/implausible), throws `SessionExpiredError` and persists nothing rather than storing a born-dead session.
- **Reads (`getSession` / `getUser`)** — once the ceiling is reached, the current-domain session is deleted and the call returns `undefined` (silent), so the app's existing not-logged-in redirect path fires transparently.
- **`getAccessToken`** — once the ceiling is reached, the session is deleted and `SessionExpiredError` is thrown **before** any cached token is served or any `/oauth/token` refresh is attempted (never refresh past the ceiling).
- **`startLinkUser` / `startUnlinkUser`** — same throwing enforcement, since both require a live session and pass the session's ID token to build the account link/unlink URL.
- **`getAccessTokenForConnection` is intentionally NOT capped** — connection (Token Vault) tokens are the upstream IdP's own tokens, governed by that IdP's `expires_in`, not the Auth0 session ceiling. They keep working past the ceiling. (Per the closed Token Vault decision; matches `auth0-server-python`.)
- **Refresh preserves the ceiling** — write-once per authentication; never re-derived from a refresh response.
- **Cookie max-age** is capped at the ceiling as a defense-in-depth backstop. Idle/absolute timeouts stay independent.
- **Fail-open**: a missing or malformed claim means "no ceiling," never "already expired." A **millisecond-scale value is rejected** (`>= 1e10`) so an Action that forgets seconds can't silently disable enforcement.
- **~30s negative leeway** for clock skew; integer-seconds comparisons throughout.
- Adds `SessionExpiredError` (`code: 'session_expired'`), types `session_expiry?: number` on the ID-token claims (`IDTokenClaims` in `@auth0/auth0-auth-js`), and a "Session expiry from upstream IdP" guide in EXAMPLES.md.

## Behavior changes for consumers (opt-in, once the feature is enabled)

No API breakage, but two new runtime behaviors surface **only** after a connection enables the option and a user logs in through it:

- `getSession()` / `getUser()` can return `undefined` for a previously-logged-in user once the ceiling passes — apps that assumed these always return a value after login should add a null check.
- `getAccessToken()`, `startLinkUser()`, and `startUnlinkUser()` can throw `SessionExpiredError` past the ceiling. (`getAccessTokenForConnection()` does **not** throw on the ceiling.)

## Public API (all additive)

- New optional field `SessionData.sessionExpiresAt?: number`
- New exported error `SessionExpiredError` (`code: 'session_expired'`)
- New exported `IDTokenClaims` type (extends `IDToken` with `session_expiry?: number`); `TokenResponse.claims` widened to it
- New optional param on the `protected calculateMaxAge(createdAt, sessionExpiresAt?)`

Nothing removed, renamed, or behaviorally retyped.

## Test plan

- [x] `extractSessionExpiry` (valid/invalid shapes, millisecond rejection at the `1e10` bound), `isSessionExpiryReached`, `isSessionExpiryInPast` (leeway boundaries, ms-`iat` fallback)
- [x] `updateStateData` is preserve-only: does not stamp on fresh login; preserves the stored ceiling across refresh **even when the response carries `session_expiry`**
- [x] `applySessionExpiryAtLogin`: stamps; clears on a no-ceiling re-login; updates on re-login; throws when born-expired
- [x] `getSession` / `getUser`: expired → undefined + delete; future/absent → unchanged; resolver-mode foreign domain not deleted
- [x] `getAccessToken`: expired → throws `SessionExpiredError`, no `/oauth/token` call, session deleted; future → cached token served (regression)
- [x] `getAccessTokenForConnection`: NOT capped — serves past the ceiling, does not throw or delete
- [x] `startLinkUser` / `startUnlinkUser`: expired → throws, URL builder never called
- [x] MFA `verify`: stamps the ceiling; throws + persists nothing when born-expired
- [x] `completeInteractiveLogin`: born-expired → throws, persists nothing, transaction still deleted
- [x] Cookie max-age capped at the ceiling (rolling + non-rolling)

## Release Notes  

* **New Features**
  * Enterprise OIDC sessions now honor the upstream `session_expiry` ceiling (`sessionExpiresAt`) and automatically stop using the session when the ceiling is reached (with clock leeway).
  * Session expiry is applied consistently during MFA verification as well.

* **Bug Fixes**
  * Expired session state is cleared across session accessors; access-token retrieval now throws `SessionExpiredError`.
  * Connection-token retrieval remains uncapped by the session ceiling.
  * Session cookie lifetime/max-age is capped to the ceiling.

* **Documentation**
  * Added examples and upgrade guidance for IPSIE `session_expiry` behavior.

* **Tests**
  * Added unit coverage for parsing, boundary/leeway logic, and lockout behavior.

### DIFF
--- a/packages/auth0-server-js/EXAMPLES.md
+++ b/packages/auth0-server-js/EXAMPLES.md
@@ -1367,16 +1367,18 @@ When you use an Okta or OIDC **enterprise connection** configured with `id_token
 No configuration or code change is required. When the claim is present, the SDK:
 
 - persists it on the session as a top-level `sessionExpiresAt` (Unix seconds);
-- treats the session as expired once `sessionExpiresAt` is reached (with a small 30-second leeway for clock skew), on **every** `getSession()` / `getUser()` / `getAccessToken()` / `getAccessTokenForConnection()` call;
+- treats the session as expired once `sessionExpiresAt` is reached (with a small 30-second leeway for clock skew), on **every** `getSession()` / `getUser()` / `getAccessToken()` call;
 - never refreshes a token once the ceiling has passed; and
 - caps the session cookie lifetime at the ceiling as a defense-in-depth backstop.
+
+`getAccessTokenForConnection()` is **not** capped by the ceiling: connection (Token Vault) tokens are the upstream IdP's own tokens, governed by that IdP's `expires_in` rather than the Auth0 session, so they keep working past the ceiling.
 
 This is layered **on top of** your existing idle and absolute session timeouts — it does not replace them. The session ends at whichever limit is reached first.
 
 ### Behavior on expiry
 
 - `getSession()` and `getUser()` return `undefined` (the session is also cleared from the store). Your existing "not logged in → redirect to login" path runs unchanged.
-- `getAccessToken()` and `getAccessTokenForConnection()` throw `SessionExpiredError` (`error.code === 'session_expired'`). Catch it and send the user to log in again.
+- `getAccessToken()` throws `SessionExpiredError` (`error.code === 'session_expired'`) — as do `startLinkUser()` and `startUnlinkUser()`. Catch it and send the user to log in again. (`getAccessTokenForConnection()` does not throw on the ceiling; see the note above.)
 
 ```ts
 import { SessionExpiredError } from '@auth0/auth0-server-js';
--- a/packages/auth0-server-js/EXAMPLES.md
+++ b/packages/auth0-server-js/EXAMPLES.md
@@ -1358,6 +1358,63 @@ const accessToken = await serverClient.getAccessTokenForConnection({}, storeOpti
 
 Read more above in [Configuring the Store](#configuring-the-store)
 
+## Session expiry from upstream IdP (IPSIE `session_expiry`)
+
+When you use an Okta or OIDC **enterprise connection** configured with `id_token_session_expiry_supported: true`, Auth0 includes a `session_expiry` claim in the ID token it issues to your application. This claim is an absolute Unix timestamp (seconds) that marks the latest moment the upstream identity provider considers the session valid. It is computed by Auth0 as the earliest of: your tenant's absolute session lifetime, the upstream IdP's own session expiry, and any value an Action sets via `api.session.setExpiresAt`.
+
+### What the SDK does automatically
+
+No configuration or code change is required. When the claim is present, the SDK:
+
+- persists it on the session as a top-level `sessionExpiresAt` (Unix seconds);
+- treats the session as expired once `sessionExpiresAt` is reached (with a small 30-second leeway for clock skew), on **every** `getSession()` / `getUser()` / `getAccessToken()` / `getAccessTokenForConnection()` call;
+- never refreshes a token once the ceiling has passed; and
+- caps the session cookie lifetime at the ceiling as a defense-in-depth backstop.
+
+This is layered **on top of** your existing idle and absolute session timeouts — it does not replace them. The session ends at whichever limit is reached first.
+
+### Behavior on expiry
+
+- `getSession()` and `getUser()` return `undefined` (the session is also cleared from the store). Your existing "not logged in → redirect to login" path runs unchanged.
+- `getAccessToken()` and `getAccessTokenForConnection()` throw `SessionExpiredError` (`error.code === 'session_expired'`). Catch it and send the user to log in again.
+
+```ts
+import { SessionExpiredError } from '@auth0/auth0-server-js';
+
+try {
+  const { accessToken } = await serverClient.getAccessToken(storeOptions);
+  // use accessToken
+} catch (error) {
+  if (error instanceof SessionExpiredError) {
+    return redirect(await serverClient.startInteractiveLogin({ authorizationParams: { prompt: 'login' } }, storeOptions));
+  }
+  throw error;
+}
+```
+
+### Reading the value (optional)
+
+To show a "your session ends soon" prompt, read the ceiling off the session:
+
+```ts
+const session = await serverClient.getSession(storeOptions);
+if (session?.sessionExpiresAt) {
+  const remainingSeconds = session.sessionExpiresAt - Math.floor(Date.now() / 1000);
+  // e.g. show a banner when remainingSeconds < 5 * 60
+}
+```
+
+Do not copy `sessionExpiresAt` into a separate long-lived store and trust it later — always re-read it relative to the current wall clock.
+
+### Upgrading existing apps
+
+Once an enterprise connection has the option enabled, `getSession()` / `getUser()` can return `undefined` for a user who was previously logged in, once the ceiling is reached. If your code assumed these always return a value after login, add a null check. Sessions created before the upgrade (or through connections without the option) have no `sessionExpiresAt` and behave exactly as before.
+
+### Prerequisites
+
+- The connection must be an `okta` or `oidc` enterprise connection with `id_token_session_expiry_supported: true` (Dashboard toggle "Use ID Token for Session Expiry", Management API, or Terraform).
+- Authorization Code flow.
+
 ## Logout
 
 Logging out ensures the stored tokens and user information are removed, and that the user is no longer considered logged-in by the SDK.
--- a/packages/auth0-server-js/src/server-client.spec.ts
+++ b/packages/auth0-server-js/src/server-client.spec.ts
@@ -6144,9 +6144,54 @@ test('completeInteractiveLogin - throws SessionExpiredError and persists nothing
 
     // State must NOT be persisted when lockout guard triggers
     expect(mockStateStore.set).not.toHaveBeenCalled();
+    // The spent transaction must still be cleaned up even though the lockout threw
+    expect(mockTransactionStore.delete).toHaveBeenCalled();
   } finally {
     getTokenByCodeSpy.mockRestore();
   }
+});
+
+test('loginBackchannel - throws SessionExpiredError and persists nothing when session_expiry is already past at login', async () => {
+  const iat = Math.floor(Date.now() / 1000);
+  // The token endpoint returns an ID token whose session_expiry is at iat → born expired.
+  const bornExpiredIdToken = await generateToken(domain, 'user_123', '<client_id>', undefined, {
+    iat,
+    session_expiry: iat,
+  });
+
+  server.use(
+    http.post(mockOpenIdConfiguration.token_endpoint, async () => {
+      return HttpResponse.json({
+        access_token: accessToken,
+        id_token: bornExpiredIdToken,
+        expires_in: 60,
+        token_type: 'Bearer',
+        scope: '<scope>',
+      });
+    })
+  );
+
+  const mockStateStore = {
+    get: vi.fn(),
+    set: vi.fn(),
+    delete: vi.fn(),
+    deleteByLogoutToken: vi.fn(),
+  };
+
+  const serverClient = new ServerClient({
+    domain,
+    clientId: '<client_id>',
+    clientSecret: '<client_secret>',
+    transactionStore: { get: vi.fn(), set: vi.fn(), delete: vi.fn() },
+    stateStore: mockStateStore,
+  });
+
+  await expect(
+    serverClient.loginBackchannel({ loginHint: { sub: '<sub>' }, bindingMessage: '<binding_message>' })
+  ).rejects.toBeInstanceOf(SessionExpiredError);
+
+  // A born-expired session must not be persisted.
+  expect(mockStateStore.set).not.toHaveBeenCalled();
 });
 
 test('getSession - returns undefined and deletes the session once the session_expiry ceiling is reached', async () => {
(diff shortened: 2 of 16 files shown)


packages/auth0-server-js/src/server-client.ts
```diff
@@ -635,6 +661,11 @@ export class ServerClient<TStoreOptions = unknown> {
       }
     }
 
+    if (stateData && isSessionExpiryReached(stateData.sessionExpiresAt)) {
+      await this.#stateStore.delete(this.#stateStoreIdentifier, storeOptions);
+      throw new SessionExpiredError();
+    }
```
kishore7snehil: The connection-token path here shouldn't be gated by the `session_expiry` ceiling at all. Connection tokens (Token Vault) follow the upstream IdP's own `expires_in`, not the Auth0 app-session ceiling. As written, this throws `SessionExpiredError` on a connection-token fetch that should have succeeded.

Could you drop this whole `if (stateData && isSessionExpiryReached(...))` block from `getAccessTokenForConnection`? The same block is correct on `getAccessToken`, `getUser`, `getSession`- it's only this connection path that should keep working past the ceiling.

packages/auth0-server-js/src/server-client.spec.ts
```diff
@@ -4338,3 +4394,287 @@ test('Telemetry - should include Auth0-Client header in token requests', async (
...
+  };
+  mockStateStore.get.mockResolvedValue(stateData);
+
+  const result = await serverClient.getAccessToken();
+
+  expect(result.accessToken).toBe('<access_token>');
+  expect(mockStateStore.delete).not.toHaveBeenCalled();
+});
+
+test('getAccessTokenForConnection - throws SessionExpiredError and never calls the token endpoint once the ceiling is reached', async () => {
```
kishore7snehil: This test pins down the behaviour we're about to remove (see the note on `getAccessTokenForConnection` in server-client.ts). Once the gate comes off, this should flip: with a ceiling in the past, `getAccessTokenForConnection` shouldn't throw - it should carry on to the connection-token path and serve or fetch the token as usual, since connection tokens track the IdP's own expiry rather than the app session.

packages/auth0-server-js/EXAMPLES.md
```diff
@@ -1054,6 +1054,63 @@ const accessToken = await serverClient.getAccessTokenForConnection({}, storeOpti
...
+## Session expiry from upstream IdP (IPSIE `session_expiry`)
+
+When you use an Okta or OIDC **enterprise connection** configured with `id_token_session_expiry_supported: true`, Auth0 includes a `session_expiry` claim in the ID token it issues to your application. This claim is an absolute Unix timestamp (seconds) that marks the latest moment the upstream identity provider considers the session valid. It is computed by Auth0 as the earliest of: your tenant's absolute session lifetime, the upstream IdP's own session expiry, and any value an Action sets via `api.session.setExpiresAt`.
+
+### What the SDK does automatically
+
+No configuration or code change is required. When the claim is present, the SDK:
+
+- persists it on the session as a top-level `sessionExpiresAt` (Unix seconds);
+- treats the session as expired once `sessionExpiresAt` is reached (with a small 30-second leeway for clock skew), on **every** `getSession()` / `getUser()` / `getAccessToken()` / `getAccessTokenForConnection()` call;
```
kishore7snehil: Small docs fix to go with the connection-token change: this line should drop `getAccessTokenForConnection()`, since the ceiling isn't enforced on that path. It should just read `getSession()` / `getUser()` / `getAccessToken()`.

packages/auth0-server-js/EXAMPLES.md
```diff
@@ -1054,6 +1054,63 @@ const accessToken = await serverClient.getAccessTokenForConnection({}, storeOpti
...
+- never refreshes a token once the ceiling has passed; and
+- caps the session cookie lifetime at the ceiling as a defense-in-depth backstop.
+
+This is layered **on top of** your existing idle and absolute session timeouts — it does not replace them. The session ends at whichever limit is reached first.
+
+### Behavior on expiry
+
+- `getSession()` and `getUser()` return `undefined` (the session is also cleared from the store). Your existing "not logged in → redirect to login" path runs unchanged.
+- `getAccessToken()` and `getAccessTokenForConnection()` throw `SessionExpiredError` (`error.code === 'session_expired'`). Catch it and send the user to log in again.
```
kishore7snehil: Same here - `getAccessTokenForConnection()` shouldn't be in the "throws SessionExpiredError" sentence. Only `getAccessToken()` throws on the ceiling (along with `startLinkUser()` and `startUnlinkUser()`); the connection-token path keeps working.

packages/auth0-server-js/src/state/utils.ts
```diff
@@ -63,15 +102,29 @@ export function updateStateData(
       refreshToken: tokenEndpointResponse.refreshToken ?? stateData.refreshToken,
       tokenSets,
       domain: context?.domain ?? stateData.domain,
+      sessionExpiresAt: extractSessionExpiry(tokenEndpointResponse.claims) ?? stateData.sessionExpiresAt,
```
kishore7snehil: There's a gap here on re-login. The check that rejects an already-expired session at login - the one that throws when session_expiry is at or before iat - only exists in the `else` branch below, the one that builds a brand-new session. This branch, which runs when the same user logs in again and we already have their state, just carries the ceiling forward on this line without that check. So if someone re-logs-in and the new ID token's session_expiry is already in the past, we persist it instead of rejecting it.

It's not a security problem, since the read-time checks still treat that session as expired the moment anyone reads it. But the requirements ask us to fail at login rather than store an already-expired session, and right now we only do that for fresh logins, not re-logins. Applying the same check here before persisting would close it.

One smaller thing on the same line, lower priority. The `?? stateData.sessionExpiresAt` fallback means that if a response ever came back carrying a later session_expiry, it would push the stored ceiling further out. That's harmless for re-login, but on the refresh path we're not supposed to re-derive the ceiling from the refresh response, because that's exactly how a session could get extended past its limit. It can't actually happen today since the refresh grant doesn't return this claim, and auth0-server-python keeps the original login value and never re-reads it on refresh. Worth a short code comment here noting we're relying on the platform not sending the claim on refresh.

packages/auth0-server-js/src/server-client.spec.ts
```diff
@@ -1633,6 +1633,62 @@ test('completeInteractiveLogin - should call cookieHandler.setCookie with custom
...
+
+  const getTokenByCodeSpy = vi.spyOn(AuthClient.prototype, 'getTokenByCode').mockResolvedValue(tokenResponse);
+
+  try {
+    await expect(
+      serverClient.completeInteractiveLogin(new URL(`https://${domain}?code=123`))
+    ).rejects.toBeInstanceOf(SessionExpiredError);
+
+    // State must NOT be persisted when lockout guard triggers
+    expect(mockStateStore.set).not.toHaveBeenCalled();
```
kishore7snehil: In `completeInteractiveLogin` the code gets exchanged for tokens first, and then `updateStateData` throws the lockout error before we reach the line that deletes the transaction - so the spent transaction just sits there until its TTL clears it. Low severity since the code can't be replayed, but` auth0-server-python` deletes the transaction before raising in the same spot.

This test only checks that state wasn't saved. If the fix goes into `completeInteractiveLogin`, it'd be good to also assert here that the transaction was deleted.

packages/auth0-server-js/src/state/utils.ts
```diff
@@ -1,5 +1,44 @@
...
+ *
+ * Fail-open by design: returns `undefined` (meaning "no ceiling") unless the value is a
+ * positive integer. A missing or malformed claim MUST NEVER be treated as an already-expired
+ * session — a platform glitch must not lock out every enterprise user.
+ *
+ * @param claims The decoded ID token claims, or undefined.
+ * @returns The ceiling in Unix seconds, or undefined when absent/invalid.
+ */
+export function extractSessionExpiry(claims: TokenResponse['claims'] | undefined): number | undefined {
+  const value = claims?.session_expiry;
```
kishore7snehil, line 45: One edge case  which I can think of is  `extractSessionExpiry` accepts any positive integer, so a value that's accidentally in milliseconds (like 1748566800000) can go through as a seconds timestamp far in the future - which quietly means the ceiling never triggers .

Also, `session_expiry` won't be emitted natively and it's set by a customer Post-Login Action so this can easily be sailed through.

packages/auth0-server-js/src/state/utils.spec.ts
```diff
@@ -536,3 +543,188 @@ test('updateStateData - should retain or override domain for existing sessions',
...
+  expect(extractSessionExpiry({ session_expiry: 1748566800 } as never)).toBe(1748566800);
+});
+
+test('extractSessionExpiry - returns undefined when the claim is absent', () => {
+  expect(extractSessionExpiry({ sub: '<sub>' } as never)).toBeUndefined();
+  expect(extractSessionExpiry(undefined)).toBeUndefined();
+});
+
+test('extractSessionExpiry - returns undefined for invalid shapes (fail-open)', () => {
+  expect(extractSessionExpiry({ session_expiry: '1748566800' } as never)).toBeUndefined(); // string
```
kishore7snehil, line 559: A minor suggestion -  the `as never` casts in these tests show session_expiry isn't on the typed claims object (`claims` is `IDToken` from openid-client). The ask is  for `session_expiry?: number` on the claims type so developers get autocomplete without casting.

packages/auth0-server-js/src/store/abstract-session-store.ts
```diff
@@ -16,15 +16,27 @@ export abstract class AbstractSessionStore<TStoreOptions> extends AbstractStateS
 
   /**
    * calculateMaxAge calculates the max age of the session based on createdAt and the rolling and absolute durations.
+   * When sessionExpiresAt is provided, caps the maxAge to not exceed the time until that ceiling.
    */
-  protected calculateMaxAge(createdAt: number) {
+  protected calculateMaxAge(createdAt: number, sessionExpiresAt?: number) {
```
kishore7snehil, line 21: The two things now do the same clamp-and-floor work, so it reads as duplicated. Since the only real difference between rolling and non-rolling is how the base maxAge is computed, we could work that out once and then do the ceiling clamp and the `> 0` floor a single time:

```ts
protected calculateMaxAge(createdAt: number, sessionExpiresAt?: number) {
  const now = (Date.now() / 1000) | 0;

  let maxAge = this.#rolling
    ? Math.min(now + this.#inactivityDuration, createdAt + this.#absoluteDuration) - now
    : this.#absoluteDuration;

  if (sessionExpiresAt !== undefined) {
    maxAge = Math.min(maxAge, sessionExpiresAt - now);
  }

  return maxAge > 0 ? maxAge : 0;
}
```

kishore7snehil (changes requested):

packages/auth0-server-js/src/server-client.ts
nandan-bhat: Addressed in 39311cdff36ba3df863f36d07fa192eee1c54293

packages/auth0-server-js/EXAMPLES.md
nandan-bhat: Addressed in 39311cdff36ba3df863f36d07fa192eee1c54293

packages/auth0-server-js/EXAMPLES.md
nandan-bhat: Addressed in 39311cdff36ba3df863f36d07fa192eee1c54293

packages/auth0-server-js/src/state/utils.ts
nandan-bhat: Addressed in 39311cdff36ba3df863f36d07fa192eee1c54293

packages/auth0-server-js/src/server-client.spec.ts
nandan-bhat: Addressed in 39311cdff36ba3df863f36d07fa192eee1c54293

packages/auth0-server-js/src/state/utils.ts
nandan-bhat: Addressed in 39311cdff36ba3df863f36d07fa192eee1c54293

packages/auth0-server-js/src/state/utils.spec.ts
nandan-bhat: Addressed in 39311cdff36ba3df863f36d07fa192eee1c54293

packages/auth0-server-js/src/store/abstract-session-store.ts
nandan-bhat: Addressed in 39311cdff36ba3df863f36d07fa192eee1c54293

packages/auth0-server-js/src/server-client.spec.ts
nandan-bhat: Addressed in 39311cdff36ba3df863f36d07fa192eee1c54293

kishore7snehil (approved):
LGTM!

PR #200, opened 2026-06-19
feat(auth0-auth-js): validate organization claim in returned ID token

nandan-bhat:
### Description

Adds organization claim validation to the `auth0-auth-js` token methods that support organizations and return a user ID token. When `organization` is provided, the returned ID token's organization claim is validated against it:
- an organization ID (the `org_` prefix) is matched **exactly** (case-sensitive) against the `org_id` claim;
- an organization name (no `org_` prefix) is matched **case-insensitively** against the `org_name` claim.

A missing or mismatched claim throws a new `OrganizationValidationError`. Validation is **opt-in** (it only runs when `organization` is provided) and only applies **when an ID token is present** — exchanges that return only an access token (e.g. M2M token exchange) are a no-op, since there is no ID-token claim to validate.
Validation is applied to the three methods the platform supports `organization` on with a user ID token:
- `AuthClient.getTokenByCode`
- `AuthClient.exchangeToken` (custom token exchange profile path)
- `PasskeyClient.getTokenByPasskey`

The validation messaging follows the SDK Organizations validation guidance.

### Changes

- Add `OrganizationValidationError` (code `organization_validation_error`, extends `Error`) — a post-exchange claim-validation failure, distinct from the OAuth token-request errors. Not rewrapped as `TokenByCodeError` / `TokenExchangeError`.
- Add a shared `validateOrganization()` helper (internal to the package) used by all three call sites.
- Add optional `organization` to `TokenByCodeOptions` (already present on the passkey and exchange option types).
- Document the behavior and `OrganizationValidationError` in `EXAMPLES.md` for all three methods.

 ### Behavior change (note for existing callers)
`exchangeToken` (custom token exchange profile path) and `PasskeyClient.getTokenByPasskey` already accepted and forwarded the `organization` parameter but the returned ID token's organization claim was **not** validated. As of this change, when `organization` is provided **and** the response includes an ID token, a missing or mismatched `org_id`/`org_name` claim now throws `OrganizationValidationError`.

- This only affects callers who **already pass `organization`** to these methods. Callers who don't pass it are unaffected.
- Exchanges that return only an access token (no ID token, e.g. M2M) are unchanged — there is no ID-token claim to validate, so validation is a no-op.
- `getTokenByCode` is new behavior in the same release (it now validates the claim when `organization` is provided).

### Testing
- [x] Unit tests for each method covering `org_id` exact match / mismatch / missing, `org_name` case-insensitive match / mismatch / missing, blank input, no-ID-token (no-op) and the no-op path when `organization` is unset.
- [x] Full `auth0-auth-js` suite green; `auth0-api-js` consumer suite green; build and lint clean.

## Release Notes

* **New Features**
  * Added organization claim validation for token exchange methods (`getTokenByCode`, `exchangeToken`, `getTokenByPasskey`) with optional `organization` parameter
  * New `OrganizationValidationError` thrown when organization claims mismatch

* **Tests**
  * Added comprehensive test coverage for organization validation scenarios

* **Documentation**
  * Updated examples documenting organization claim validation behavior and matching rules

### DIFF
--- a/packages/auth0-auth-js/src/auth-client.spec.ts
+++ b/packages/auth0-auth-js/src/auth-client.spec.ts
@@ -1402,6 +1402,25 @@ describe('getTokenByCode - organization validation', () => {
     expect(res.accessToken).toBe(accessToken);
   });
 
+  test('skips validation when org is requested but no ID token is returned', async () => {
+    server.use(
+      http.post(mockOpenIdConfiguration.token_endpoint, async () => {
+        return HttpResponse.json({
+          access_token: accessToken,
+          expires_in: 60,
+          token_type: 'Bearer',
+          scope: '<scope>',
+        });
+      })
+    );
+    const authClient = newClient();
+    const res = await authClient.getTokenByCode(new URL(`https://${domain}?code=123`), {
+      codeVerifier: '123',
+      organization: 'org_abc123',
+    });
+    expect(res.accessToken).toBe(accessToken);
+  });
+
   test('throws a clear error when organization is only whitespace', async () => {
     useOrgTokenHandler({ org_name: 'acme-corp' });
     const authClient = newClient();
--- a/packages/auth0-auth-js/src/auth-client.spec.ts
+++ b/packages/auth0-auth-js/src/auth-client.spec.ts
@@ -2,7 +2,7 @@ import { expect, test, afterAll, beforeAll, beforeEach, vi, afterEach, describe 
 import { setupServer } from 'msw/node';
 import { http, HttpResponse } from 'msw';
 import { AuthClient } from './auth-client.js';
-import { NotSupportedError, isMfaRequiredError, TokenByPasswordError } from './errors.js';
+import { NotSupportedError, isMfaRequiredError, TokenByPasswordError, OrganizationValidationError } from './errors.js';
 import { PasskeyGetTokenError } from './passkey/errors.js';
 import { ExchangeProfileOptions } from './types.js';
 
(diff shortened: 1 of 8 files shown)


packages/auth0-auth-js/src/auth-client.ts
```diff
@@ -829,6 +830,11 @@ export class AuthClient {
       );
 
       const tokenResponse = TokenResponse.fromTokenEndpointResponse(tokenEndpointResponse);
+
+      if (options.organization) {
+        validateOrganization(tokenResponse.claims, options.organization);
```
kishore7snehil: This validate call sits inside the `try`, so the `catch` right below has to special-case `OrganizationValidationError` and re-throw it (line 851) to stop it being wrapped as a `TokenExchangeError`. The passkey method runs the same validation just *after* its try/catch (), which skips that thing and the validation error propagates on its own.

Shall we move this call to after the try/catch here? The only throwing call in the block is the grant request; building the response and the actorToken decode don't need to sit inside the org check, so moving it down lets the `instanceof` guard go away.

packages/auth0-auth-js/src/auth-client.ts
```diff
@@ -952,8 +967,17 @@ export class AuthClient {
         pkceCodeVerifier: options.codeVerifier,
       });
 
-      return TokenResponse.fromTokenEndpointResponse(tokenEndpointResponse);
+      const tokenResponse = TokenResponse.fromTokenEndpointResponse(tokenEndpointResponse);
+
+      if (options.organization) {
+        validateOrganization(tokenResponse.claims, options.organization);
```
kishore7snehil: Same as the exchange path above - moving this to after the try/catch drops the need for the `instanceof OrganizationValidationError` re-throw on line 978.

packages/auth0-auth-js/src/auth-client.spec.ts
```diff
@@ -2674,6 +2806,75 @@ describe('exchangeToken', () => {
     // Verify that allowed custom params are still forwarded (denylist is selective)
     expect(capturedCustomParam).toBe('allowed');
   });
+
+  describe('organization validation', () => {
```
kishore7snehil, line 2821: The `getTokenByCode` block above covers `org_name` mismatch and the whitespace-only input case, this exchange block has the `org_id` mismatch and missing cases but not those two. 

They all route through the same helper so it isn't a correctness gap, but mirroring the two missing cases here would keep the three method suites symmetric.

packages/auth0-auth-js/src/passkey/passkey-client.spec.ts
```diff
@@ -1119,6 +1130,91 @@ describe('PasskeyClient', () => {
         expect(error.cause?.mfa_token).toBeUndefined();
       }
     });
+
+    describe('organization validation', () => {
```
kishore7snehil, line 1134: Same as the exchange suite -the `org_name` mismatch and whitespace-only cases aren't here. Worth mirroring so all three read the same.

packages/auth0-auth-js/src/utils.ts
```diff
@@ -7,3 +10,52 @@ export function stripUndefinedProperties<T extends object>(value: T): Partial<T>
...
+ * for example, token-exchange flows that return only an access token. When an ID token
+ * is present, a missing or mismatched organization claim throws.
+ *
+ * @throws {OrganizationValidationError} when `organization` is blank, or when an ID
+ * token is present and its organization claim is missing or does not match.
+ */
+export function validateOrganization(claims: IDToken | undefined, organization: string): void {
+  const org = organization.trim();
+
+  if (!org) {
```
kishore7snehil: Two small things about the blank check here. 

First, a whitespace-only `organization` throws "must not be blank", but an empty string never reaches this line - `if (options.organization)` at the call sites is falsy for `''`, so it silently skips validation. Both read as "blank" yet behave differently. 

Second, because this runs after the exchange, a blank value only fails once the token request has already gone out and come back; the subject-token checks in `exchangeToken` fail before any network call, so the input-shape part of this could move up front to match.

kishore7snehil (changes requested):

packages/auth0-auth-js/src/auth-client.ts
nandan-bhat: Addressed in f28d1652246214d1c68e7847b285e6a6bcd4f598

packages/auth0-auth-js/src/auth-client.ts
nandan-bhat: Addressed in f28d1652246214d1c68e7847b285e6a6bcd4f598

packages/auth0-auth-js/src/auth-client.spec.ts
nandan-bhat: Addressed in f28d1652246214d1c68e7847b285e6a6bcd4f598

packages/auth0-auth-js/src/passkey/passkey-client.spec.ts
nandan-bhat: Addressed in f28d1652246214d1c68e7847b285e6a6bcd4f598

packages/auth0-auth-js/src/utils.ts
nandan-bhat: Addressed in f28d1652246214d1c68e7847b285e6a6bcd4f598

kishore7snehil (approved):
LGTM!

PR #204, opened 2026-06-23
refactor: Moved passkeys API to a new passkey client

pmathew92:
### Summary

  Refactors the flat passkey methods on ServerClient into a dedicated passkey sub-client. This groups the passkey surface under one namespace and aligns the public API shape

  #### Before
  ```
  await serverClient.passkeyRegister({ email, name });
  await serverClient.passkeyChallenge();
  await serverClient.passkeyGetToken({ authSession, credential });
  ```
  
  #### After
  ```
  await serverClient.passkey.register({ email, name });
  await serverClient.passkey.challenge();
  await serverClient.passkey.getToken({ authSession, credential });
  ```

  ### Changes

  - New `src/passkey/ module: ServerPasskeyClient (register / challenge / getToken)`, its options interface, and a barrel that also re-exports the PasskeyRegisterError / PasskeyChallengeError / PasskeyGetTokenError classes for
  instanceof narrowing.
  - ServerClient: removed the three flat methods; added a get passkey() accessor. The sub-client is wired with `resolveDomain / getAuthClient` functions rather than a frozen domain.
  - Resolver-mode support preserved: unlike mfa (static-domain only, getter throws in resolver mode), passkey resolves the domain per-call, so it works in both static and multi-tenant (resolver) modes and is always available.
  - utils.ts: hoisted DEFAULT_SCOPES / ensureOpenIdScope so both ServerClient and the passkey sub-client share one implementation.
  - Exports: passkey barrel added to index.ts; care taken to avoid ESM export * name collisions (type aliases stay in types.ts; the barrel only exports the new client + error classes).
  - Docs: EXAMPLES.md updated to the passkey.* API, plus a resolver-mode note that the same storeOptions must be passed to register()/challenge() and getToken(), since authSession is bound to the issuing tenant.

 
  Test plan
  
  - npm test — passkey suite migrated to the sub-client (10 tests incl. a new resolver-mode domain-resolution test); 248 pass. (2 pre-existing Telemetry ESM-spy failures are unrelated/out of scope.)
  - npm run build — generated d.ts exposes get passkey(), ServerPasskeyClient, PasskeyGetTokenResult, all 3 error classes, and all passkey type aliases; no duplicate exports.
  - npm run lint — clean.
  - Manual testing on passjey signup and login

### Checklist

- [X] I have added documentation for new/changed functionality in this PR or in auth0.com/docs
- [X] All active GitHub checks for tests, formatting, and security are passing
- [X] The correct base branch is being used, if not the default branch

## Summary by CodeRabbit

* **New Features**
  * Introduced a dedicated server-side passkey client with `register()`, `challenge()`, and `getToken()` flow orchestration.
  * Updated passkey API usage to the newer `passkey.register()`, `passkey.challenge()`, and `passkey.getToken()` method names.
* **Documentation**
  * Refreshed passkey and Passkeys+MCD examples, including improved callouts and mTLS guidance formatting.
  * Clarified MFA handling and state persistence behavior across the passkey flow.
* **Bug Fixes / Behavior Changes**
  * `getToken()` now ensures `openid` is present (without duplication) when exchanging scopes.
* **Tests**
  * Updated server client passkey tests to match the new passkey API and verify state-store write behavior.

### DIFF
--- a/packages/auth0-server-js/src/passkey/index.ts
+++ b/packages/auth0-server-js/src/passkey/index.ts
@@ -2,5 +2,12 @@ export { ServerPasskeyClient } from './server-passkey-client.js';
 
 // Re-export passkey error classes from auth0-auth-js for convenience, so consumers
 // can narrow thrown errors via `instanceof` without importing auth0-auth-js directly.
+// `OrganizationValidationError` is thrown by `getToken()` when an `organization` is
+// passed and the returned ID token's organization claim is missing or mismatched.
 // The passkey option/response types are re-exported from `../types.js`.
-export { PasskeyRegisterError, PasskeyChallengeError, PasskeyGetTokenError } from '@auth0/auth0-auth-js';
+export {
+  PasskeyRegisterError,
+  PasskeyChallengeError,
+  PasskeyGetTokenError,
+  OrganizationValidationError,
+} from '@auth0/auth0-auth-js';
--- a/packages/auth0-server-js/src/passkey/index.ts
+++ b/packages/auth0-server-js/src/passkey/index.ts
@@ -1,0 +1,6 @@
+export { ServerPasskeyClient } from './server-passkey-client.js';
+
+// Re-export passkey error classes from auth0-auth-js for convenience, so consumers
+// can narrow thrown errors via `instanceof` without importing auth0-auth-js directly.
+// The passkey option/response types are re-exported from `../types.js`.
+export { PasskeyRegisterError, PasskeyChallengeError, PasskeyGetTokenError } from '@auth0/auth0-auth-js';
--- a/packages/auth0-server-js/src/passkey/server-passkey-client.ts
+++ b/packages/auth0-server-js/src/passkey/server-passkey-client.ts
@@ -1,0 +1,113 @@
+import { updateStateData } from '../state/utils.js';
+import { ensureOpenIdScope } from '../utils.js';
+import type { ServerPasskeyClientOptions } from './types.js';
+import type {
+  PasskeyRegisterOptions,
+  PasskeyRegisterResponse,
+  PasskeyChallengeOptions,
+  PasskeyChallengeResponse,
+  PasskeyGetTokenOptions,
+  PasskeyGetTokenResult,
+} from '../types.js';
+
+export class ServerPasskeyClient<TStoreOptions = unknown> {
+  readonly #options: ServerPasskeyClientOptions<TStoreOptions>;
+
+  /**
+   * @internal
+   */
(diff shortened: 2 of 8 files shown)


packages/auth0-server-js/src/passkey/index.ts
```diff
@@ -0,0 +1,6 @@
+export { ServerPasskeyClient } from './server-passkey-client.js';
+
+// Re-export passkey error classes from auth0-auth-js for convenience, so consumers
+// can narrow thrown errors via `instanceof` without importing auth0-auth-js directly.
+// The passkey option/response types are re-exported from `../types.js`.
+export { PasskeyRegisterError, PasskeyChallengeError, PasskeyGetTokenError } from '@auth0/auth0-auth-js';
```
nandan-bhat: After the merge from main, org claim validation now runs inside the passkey token exchange (#200 added it to `getTokenByPasskey`). That means `getToken()` can throw an `OrganizationValidationError` when you pass an organization. But that error is not re-exported from `auth0-server-js`, so a consumer cannot do `instanceof` on it without reaching into `auth0-auth-js` directly.

Since this barrel already re-exports the other passkey error classes for exactly this reason, can we add `OrganizationValidationError` here too ? It will keep the narrowing story consistent.

packages/auth0-server-js/src/passkey/server-passkey-client.ts
```diff
@@ -0,0 +1,113 @@
...
+   * returned by `register()` / `challenge()` together with the serialized credential.
+   *
+   * In resolver (multi-tenant) mode, pass the same `storeOptions` you passed to
+   * `register()` / `challenge()` so the token exchange resolves the same tenant
+   * that issued the `authSession`; otherwise the exchange will fail.
+   *
+   * @param options The auth session, serialized credential, and optional realm/scope/audience/organization.
+   * @param storeOptions Optional options used to pass to the State Store (and to resolve the domain in resolver mode).
+   *
+   * @throws {PasskeyGetTokenError} If there was an issue exchanging the credential for tokens. When the cause is `mfa_required`, use `isMfaRequiredError(error)` to narrow the error and read `cause.mfa_token`. No session is persisted in this case.
```
nandan-bhat, line 86: Small follow-up to the same point: this `JSDoc` only mentions `PasskeyGetTokenError`, but after the main merge `getToken()` can also throw `OrganizationValidationError` when an organization is passed and the returned `id_token` does not match. Can we add a `@throws` line for it so the behavior is documented ? A short test that asserts it propagates and no session is persisted (like the `mfa_required` one) would be a nice addition too.

nandan-bhat (approved):

PR #211, opened 2026-07-02
feat(auth0-server-js): add organization support

nandan-bhat:
This adds first class support for Auth0 Organizations in the interactive login flow.

### What changed
You can now pass an `organization` when logging in, either as a client wide default or per login:

```ts
  // Client default
  const client = new ServerClient({ /* ... */, organization: 'org_abc123' });

  // Per login (overrides the default)
  await client.startInteractiveLogin({ organization: 'org_abc123' });
 ```

The value can be an organization ID (org_abc123) or an organization name (acme-corp). It still works if you pass it through authorizationParams like before.

There is also a new invitation option for handling organization invitation links:
```ts
  await client.startInteractiveLogin({ organization: 'org_abc123', invitation: 'inv_ticket_789' });
```

When an organization is provided, the returned ID token is checked at `completeInteractiveLogin`. An `org_ prefixed` value is matched exactly against the `org_id` claim, otherwise the value is matched (case insensitive) against the `org_name` claim. If the claim is missing or does not match, it throws `OrganizationValidationError` and no session is saved.

### Notes

The actual claim check lives in `auth0-auth-js`. This SDK resolves the organization, forwards it to `/authorize`, remembers it across the redirect, and passes it to the core for validation. `OrganizationValidationError` is now re-exported so you can catch it. All new fields are optional so nothing breaks for existing apps.

### Testing

  - Added tests for organization forwarding, precedence, invitation, and claim validation at callback
  - build, lint, and the full test suite pass

## Summary by CodeRabbit

* **New Features**
  * Added organization-aware interactive login, including client-wide defaults, per-login overrides, and invitation-based flows.
  * Propagates organization details through the redirect flow to support consistent organization claim validation.
* **Bug Fixes**
  * Improved handling by rejecting blank organizations, enforcing matching organization claims (case-insensitive), and preventing session persistence when validation fails or claims are missing.
* **Documentation**
  * Expanded examples and guidance for logging in to an organization, including invitation forwarding and validation error behavior.

### DIFF
--- a/packages/auth0-server-js/src/server-client.spec.ts
+++ b/packages/auth0-server-js/src/server-client.spec.ts
@@ -6517,6 +6517,40 @@ describe('organization support', () => {
     expect(url.searchParams.get('organization')).toBe('org_viaparams');
   });
 
+  test('startInteractiveLogin - organization via client-level authorizationParams is resolved and stored', async () => {
+    const mockTransactionStore = { get: vi.fn(), set: vi.fn(), delete: vi.fn() };
+    const serverClient = new ServerClient({
+      domain,
+      clientId: '<client_id>',
+      clientSecret: '<client_secret>',
+      stateStore: new DefaultStateStore({ secret: '<secret>' }),
+      transactionStore: mockTransactionStore,
+      authorizationParams: { redirect_uri: '/test_redirect_uri', organization: 'org_clientparams' },
+    });
+
+    const url = await serverClient.startInteractiveLogin();
+
+    expect(url.searchParams.get('organization')).toBe('org_clientparams');
+    expect(mockTransactionStore.set.mock.calls[0]?.[1]?.organization).toBe('org_clientparams');
+  });
+
+  test('startInteractiveLogin - throws OrganizationValidationError on a blank organization and stores nothing', async () => {
+    const mockTransactionStore = { get: vi.fn(), set: vi.fn(), delete: vi.fn() };
+    const serverClient = new ServerClient({
+      domain,
+      clientId: '<client_id>',
+      clientSecret: '<client_secret>',
+      stateStore: new DefaultStateStore({ secret: '<secret>' }),
+      transactionStore: mockTransactionStore,
+      authorizationParams: { redirect_uri: '/test_redirect_uri' },
+    });
+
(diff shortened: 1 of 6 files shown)


packages/auth0-server-js/src/server-client.ts
```diff
@@ -275,6 +308,8 @@ export class ServerClient<TStoreOptions = unknown> {
         ...options?.authorizationParams,
         redirect_uri: redirectUri,
         scope,
+        ...(resolvedOrganization ? { organization: resolvedOrganization } : {}),
+        ...(options?.invitation ? { invitation: options.invitation } : {}),
       },
```
cschetan77, line 326: I believe, `invitation` ticket is meaningless without an organization, a guard for it can be applied as well. (And also the typescript doc for `StartInteractiveLoginOptions` interface). 
if that's not the case plz ignore.
```js
if (options?.invitation && !resolvedOrganization) {}
```

packages/auth0-server-js/src/server-client.spec.ts
```diff
@@ -6483,3 +6483,164 @@ test('startUnlinkUser - throws SessionExpiredError and never builds the unlink U
...
+      domain,
+      clientId: '<client_id>',
+      clientSecret: '<client_secret>',
+      stateStore: mockStateStore,
+      transactionStore: mockTransactionStore,
+    });
+    return { serverClient, mockStateStore, mockTransactionStore };
+  };
+
+  test('completeInteractiveLogin - succeeds when the org_id claim matches the requested organization', async () => {
```
cschetan77, line 6646: NIT: Would be good to add `org_name` validation path as well.

packages/auth0-server-js/src/server-client.ts
nandan-bhat: Addressed in 7c3d006fb3d17d09e715b955598452c9ca4610cb

packages/auth0-server-js/src/server-client.spec.ts
nandan-bhat: Addressed in 7c3d006fb3d17d09e715b955598452c9ca4610cb

cschetan77 (dismissed):

Piyush-85 (approved):

PR #214, opened 2026-07-06
feat(server-js): add revokeRefreshToken and logout revocation support

jd3vi1:
## Summary

Adds token revocation support to `auth0-server-js` so a logout or explicit revoke actually invalidates the refresh token at Auth0, not just clears the local session.

- **`ServerClient.revokeRefreshToken(options?, storeOptions?)`** -- new public method that calls `POST /oauth/revoke`. Reads the refresh token from the session store by default; accepts an explicit token via `options.token` to bypass the session lookup.
- **`logout({ revokeRefreshToken: true })`** -- convenience option that revokes before clearing the session. Best-effort: revocation failure never blocks logout.
- **Resolver mode** -- in resolver mode, the session's stored `domain` is used for the revocation request (preferred over calling the domain resolver), matching the behavior of the rest of the SDK.

## Changes

- `packages/auth0-auth-js/src/auth-client.ts` -- added `revokeToken()` method calling `POST /oauth/revoke`
- `packages/auth0-auth-js/src/types.ts` -- added `RevokeTokenOptions`; fixed JSDoc citation to RFC 7009 only
- `packages/auth0-server-js/src/server-client.ts` -- added `revokeRefreshToken()` and `revokeRefreshToken: true` option on `logout()`
- `packages/auth0-server-js/src/types.ts` -- added `RevokeRefreshTokenOptions` and `LogoutOptions.revokeRefreshToken`
- `packages/auth0-server-js/src/errors.ts` -- added `TokenRevocationError`
- `packages/auth0-server-js/EXAMPLES.md` -- added "Revoking a Refresh Token" section covering all three usage patterns
- `examples/example-express-web` -- wired `POST /auth/revoke` and `GET /auth/logout-and-revoke` routes into the existing Express example

## Test plan

- [x] `packages/auth0-auth-js`: 360 tests pass (`npm test`)
- [x] `packages/auth0-server-js`: 370 tests pass (`npm test`)
- [x] New test coverage: `revokeToken` success, error, client credentials; `revokeRefreshToken` from session, explicit token, missing token, resolver mode; `logout({ revokeRefreshToken: true })` success, no-op, failure swallowed, resolver mode via logout

## Summary by CodeRabbit

- **New Features**
  - Added support for revoking access and refresh tokens.
  - Added server-side refresh-token revocation, including optional explicit tokens.
  - Logout now attempts best-effort refresh-token revocation before completing.
  - Added refresh-token revocation to the Express example application.

- **Bug Fixes**
  - Added clear errors when no refresh token is available.
  - Logout continues successfully if token revocation fails.

- **Documentation**
  - Documented token revocation requirements, behavior, and usage.

### DIFF
--- a/packages/auth0-auth-js/src/auth-client.spec.ts
+++ b/packages/auth0-auth-js/src/auth-client.spec.ts
@@ -3992,3 +3992,92 @@ test('database.changePassword request carries the Auth0-Client telemetry header'
   await c.database.changePassword({ email: 'a@b.com', connection: 'db' });
   expect(headers?.get('Auth0-Client')).toBeTruthy();
 });
+
+describe('revokeToken', () => {
+  const revocationEndpoint = `https://${domain}/oauth/revoke`;
+
+  const setupRevocationHandlers = (handler: Parameters<typeof http.post>[1]) => {
+    server.use(
+      http.get(`https://${domain}/.well-known/openid-configuration`, () =>
+        HttpResponse.json({ ...buildOpenIdConfiguration(domain), revocation_endpoint: revocationEndpoint })
+      ),
+      http.post(revocationEndpoint, handler)
+    );
+  };
+
+  const makeClient = () =>
+    new AuthClient({
+      domain,
+      clientId: '<client_id>',
+      clientSecret: '<client_secret>',
+      discoveryCache: { ttl: 0 },
+    });
+
+  test('should successfully revoke a token', async () => {
+    let capturedToken: string | null = null;
+    let capturedHint: string | null = null;
+    setupRevocationHandlers(async ({ request }) => {
+      const body = await request.formData();
+      capturedToken = body.get('token') as string;
+      capturedHint = body.get('token_type_hint') as string;
+      return new HttpResponse(null, { status: 200 });
+    });
+
+    await expect(
+      makeClient().revokeToken({ token: '<refresh_token>', tokenTypeHint: 'refresh_token' })
+    ).resolves.toBeUndefined();
+    expect(capturedToken).toBe('<refresh_token>');
+    expect(capturedHint).toBe('refresh_token');
+  });
+
+  test('should revoke a token without tokenTypeHint', async () => {
+    let capturedHint: FormDataEntryValue | null = null;
+    setupRevocationHandlers(async ({ request }) => {
+      const body = await request.formData();
+      capturedHint = body.get('token_type_hint');
+      return new HttpResponse(null, { status: 200 });
+    });
+
+    await expect(
+      makeClient().revokeToken({ token: '<refresh_token>' })
+    ).resolves.toBeUndefined();
+    expect(capturedHint).toBeNull();
+  });
+
+  test('should throw TokenRevocationError when revocation fails', async () => {
+    setupRevocationHandlers(() =>
+      HttpResponse.json(
+        { error: '<error_code>', error_description: '<error_description>' },
+        { status: 400 }
+      )
+    );
+
+    await expect(
+      makeClient().revokeToken({ token: '<invalid_token>' })
+    ).rejects.toThrowError(
+      expect.objectContaining({
+        code: 'token_revocation_error',
+        message: 'An error occurred while trying to revoke the token.',
+        cause: expect.objectContaining({
+          error: '<error_code>',
+          error_description: '<error_description>',
+        }),
+      })
+    );
+  });
+
+  test('should send client credentials on the revocation request', async () => {
+    let capturedClientId: string | null = null;
+    let capturedClientSecret: string | null = null;
+    setupRevocationHandlers(async ({ request }) => {
+      const body = await request.formData();
+      capturedClientId = body.get('client_id') as string;
+      capturedClientSecret = body.get('client_secret') as string;
+      return new HttpResponse(null, { status: 200 });
+    });
+
+    await makeClient().revokeToken({ token: '<refresh_token>', tokenTypeHint: 'refresh_token' });
+    expect(capturedClientId).toBe('<client_id>');
+    expect(capturedClientSecret).toBe('<client_secret>');
+  });
+});
--- a/packages/auth0-auth-js/src/auth-client.spec.ts
+++ b/packages/auth0-auth-js/src/auth-client.spec.ts
@@ -4077,6 +4077,7 @@ describe('revokeToken', () => {
     });
 
     await makeClient().revokeToken({ token: '<refresh_token>', tokenTypeHint: 'refresh_token' });
+
     expect(capturedClientId).toBe('<client_id>');
     expect(capturedClientSecret).toBe('<client_secret>');
   });
--- a/packages/auth0-server-js/EXAMPLES.md
+++ b/packages/auth0-server-js/EXAMPLES.md
@@ -1506,13 +1506,12 @@ await serverClient.revokeRefreshToken({ token: '<refresh_token>' });
 
 ### Revoking on logout
 
-Pass `revokeRefreshToken: true` to `logout()` to revoke the session's refresh token as part of the logout flow.
+`logout()` automatically revokes the session's refresh token before clearing the local session.
 Revocation is best-effort: if it fails for any reason (network error, token already revoked, misconfiguration), logout still proceeds. In resolver mode, both revocation and local session deletion only occur when the stored session domain matches the resolved domain — if they differ, the session belongs to a different tenant and is left untouched.
 
 ```ts
 const logoutUrl = await serverClient.logout({
   returnTo: 'http://localhost:3000',
-  revokeRefreshToken: true,
 });
 // Redirect user to logoutUrl
 ```
--- a/packages/auth0-server-js/EXAMPLES.md
+++ b/packages/auth0-server-js/EXAMPLES.md
@@ -1507,7 +1507,7 @@ await serverClient.revokeRefreshToken({ token: '<refresh_token>' });
 ### Revoking on logout
 
 Pass `revokeRefreshToken: true` to `logout()` to revoke the session's refresh token as part of the logout flow.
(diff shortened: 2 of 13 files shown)


packages/auth0-server-js/src/server-client.ts
```diff
@@ -1051,13 +1052,58 @@ export class ServerClient<TStoreOptions = unknown> {
...
+  /**
+   * Revokes the refresh token stored in the current session, or an explicitly supplied token.
+   *
+   * @param options Optionally supply a token to revoke instead of reading from the session.
+   * @param storeOptions Optional options passed to the StateStore.
+   *
+   * @throws {MissingSessionError} If no refresh token is found in the session and none was provided.
+   * @throws {TokenRevocationError} If the revocation request fails.
+   */
+  public async revokeRefreshToken(
```
nandan-bhat, line 1090: The ticket asks for API docs and sample-app wiring as part of this work, but I don't see either in the diff. `revokeRefreshToken()` is a new public method and `EXAMPLES.md` still only documents `logout()`. Can we add an `EXAMPLES.md` entry for both the explicit token form and the session form, plus a note about `logout({ revokeRefreshToken: true })`? If docs are meant to land in a follow-up PR, can you link it here so we don't lose track of it?

packages/auth0-server-js/src/server-client.ts
```diff
@@ -1051,13 +1052,58 @@ export class ServerClient<TStoreOptions = unknown> {
...
+    options: RevokeRefreshTokenOptions = {},
+    storeOptions?: TStoreOptions
+  ): Promise<void> {
+    let refreshToken = options.token;
+
+    const stateData = await this.#stateStore.get(this.#stateStoreIdentifier, storeOptions);
```
nandan-bhat: Small thing: when the caller passes `options.token` and we are not in resolver mode, we still do this `get()` on the state store but never use the result. For stateful stores that is an extra round trip we don't need. Could skip the read when a token is supplied and we are not in resolver mode. Not blocking.

packages/auth0-server-js/src/server-client.ts
```diff
@@ -1051,13 +1052,58 @@ export class ServerClient<TStoreOptions = unknown> {
...
    * @param options Options used to configure the logout process.
    * @param storeOptions Optional options used to pass to the Transaction and State Store.
    * @returns {URL}
    */
   public async logout(options: LogoutOptions, storeOptions?: TStoreOptions) {
+    if (options.revokeRefreshToken) {
+      try {
+        await this.revokeRefreshToken({}, storeOptions);
+      } catch {
+        // best-effort: revocation failure must not block logout
```
nandan-bhat, line 1135: This catch swallows everything, which matches the best-effort behavior we want, so no objection. One downside is that a real misconfiguration during logout goes completely silent. If we have a debug or logging hook somewhere, it might be worth emitting the error here so operators can tell revocation is failing.

packages/auth0-server-js/src/server-client.spec.ts
```diff
@@ -6700,3 +6702,318 @@ describe('organization support', () => {
...
+      delete: vi.fn(),
+      deleteByLogoutToken: vi.fn(),
+    };
+
+    const stateData: StateData = {
+      user: { sub: '<sub>' },
+      idToken: '<id_token>',
+      refreshToken: '<refresh_token>',
+      tokenSets: [],
+      domain: 'other.local',
```
nandan-bhat: This test does not really prove the session domain is used. The resolver returns `other.local` and the session `domain` is also `other.local`, so the test passes whether we read the domain from the session or fall back to the resolver. Could we make the resolver return a different domain (say `resolver.local`), register the revoke handler only on `other.local`, and assert that endpoint was hit? That would actually lock in the session-domain preference.

packages/auth0-server-js/src/server-client.spec.ts
```diff
@@ -6700,3 +6702,318 @@ describe('organization support', () => {
...
+    });
+
+    const mockStateStore = {
+      get: vi.fn(),
+      set: vi.fn(),
+      delete: vi.fn(),
+      deleteByLogoutToken: vi.fn(),
+    };
+
+    mockStateStore.get.mockResolvedValue(null);
```
nandan-bhat: Because the session is `null` here, revocation would be skipped regardless of the flag, so this test would still pass even if the `revokeRefreshToken` guard in `logout()` was removed. Could we give it a full session with a `refreshToken`, call `logout({ returnTo })` without the flag, and then assert `revocationCalled` is false? That way it actually tests the opt-in gate.

packages/auth0-auth-js/src/auth-client.spec.ts
```diff
@@ -3963,3 +3963,77 @@ describe('passwordless public export surface', () => {
     expect(mod.PasswordlessVerifyError).toBeDefined();
   });
 });
+
+describe('revokeToken', () => {
```
nandan-bhat, line 3996: The ticket mentions confidential client auth (`client_secret` / `client_assertion`) as a requirement. It works today because we reuse the same `Configuration` from `#discover()`, but there is no test asserting the client credentials are actually sent on `/oauth/revoke`, like we do for token exchange. Could we add one so this guarantee does not silently regress?

packages/auth0-server-js/src/server-client.ts
jd3vi1: Done — added EXAMPLES.md with a dedicated 'Revoking a Refresh Token' section covering the three cases (session token, explicit token, revoking on logout). The section is linked from the TOC.

packages/auth0-server-js/src/server-client.ts
jd3vi1: Fixed. The store read is now gated on whether it's actually needed: we only call `stateStore.get()` when `options.token` is not provided or when in resolver mode (to get the session domain). When an explicit token is passed in static-domain mode, the store is not read at all.

packages/auth0-server-js/src/server-client.ts
jd3vi1: Acknowledged. Updated the comment in the catch block to explain the tradeoff explicitly: revocation failure is intentionally swallowed so it never blocks logout, and a logging hook would be the right place to surface it (one does not exist in the SDK today).

packages/auth0-server-js/src/server-client.spec.ts
jd3vi1: Fixed. The test now sets the domainResolver to return `resolver.local` while the session's `domain` field is `session.local`. The MSW handlers are registered only on `session.local`, so the test fails (no handler hit) if the resolver is used instead of the session domain. Added assertions for both: the revoke endpoint on `session.local` was called, and `domainResolver` was not called.

packages/auth0-server-js/src/server-client.spec.ts
jd3vi1: Fixed. The test now provides a full stateData with a `refreshToken` field in the mock store. This means the token is available and the guard is exercised — the test only passes because `revokeRefreshToken: false` (option absent) prevents the call, not because the session was empty.

packages/auth0-auth-js/src/auth-client.spec.ts
jd3vi1: Done. Added a test 'should send client credentials on the revocation request' that captures the `client_id` and `client_secret` from the revocation request body and asserts they match the values configured on the client.

packages/auth0-server-js/EXAMPLES.md
```diff
@@ -1473,6 +1477,46 @@ Once an enterprise connection has the option enabled, `getSession()` / `getUser(
...
+A specific token can be passed via `options.token`, bypassing the session lookup:
+
+```ts
+await serverClient.revokeRefreshToken({ token: '<refresh_token>' });
+```
+
+### Revoking on logout
+
+Pass `revokeRefreshToken: true` to `logout()` to revoke the session's refresh token as part of the logout flow.
+Revocation is best-effort: if it fails for any reason (network error, token already revoked, misconfiguration), logout still proceeds and the session is cleared.
```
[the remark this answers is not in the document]
jd3vi1: Updated the wording in the 'Revoking on logout' section to narrow the guarantee: revocation is described as best-effort (failure never blocks logout), and the resolver-mode session cleanup is scoped correctly — the session is only cleared when the stored domain matches the resolved domain.

packages/auth0-server-js/src/server-client.ts
jd3vi1: added example

packages/auth0-server-js/src/server-client.ts
jd3vi1: fixed with conditional needsStateData guard

packages/auth0-server-js/src/server-client.ts
jd3vi1: I couldn't find any existing logging mechanism in this package/repo, so skipping it for now

packages/auth0-server-js/src/server-client.spec.ts
jd3vi1: test rewritten with distinct resolver.local vs session.local

packages/auth0-server-js/src/server-client.spec.ts
jd3vi1: fixed by giving the mock store a session with a refresh token

packages/auth0-auth-js/src/auth-client.spec.ts
jd3vi1: added 'should send client credentials on the revocation request'

nandan-bhat (dismissed):

Piyush-85 (dismissed):
LGTM

Piyush-85 (approved):

PR #215, opened 2026-07-12
feat(auth0-server-js): add Session Transfer Token support for CTE imp…

nandan-bhat:
## What this does

This adds `Session Transfer Token` (STT) support to `auth0-server-js` for Custom Token Exchange impersonation. It lets a support or admin app log an agent into a target web app `as a customer`, so the agent can see and reproduce the customer's exact experience without knowing their password. The agent is recorded in the `act` claim, so every
impersonation is auditable.

There are two roles in the flow. The `initiator` (your app, using this SDK) requests a short lived, single use STT and redirects the agent's browser to the target app. The `target` (the customer's app) forwards the token to `/authorize`, where Auth0 redeems it and starts a session as the customer.

## What was added

Two new methods on `ServerClient`:
- `requestSessionTransferToken(options, storeOptions)` mints the STT by running a token exchange against the `urn:{domain}:session_transfer` audience. The audience is built from the resolved request domain, so it stays correct when multiple custom domains are in use.
- `buildSessionTransferRedirect(targetLoginUrl, result, options)` builds the redirect URL that carries the STT (and an optional `organization`) to the target app's login URL.

New types and error codes:

- `SessionTransferTokenResult`, `RequestSessionTransferTokenOptions`, `SessionTransferActor`,  and `BuildSessionTransferRedirectOptions`.
- A `TokenExchangeErrorCode` constant with `actor_unavailable`, `setactor_required`, and `session_transfer_disabled`.

## Key behavior

- The `actor` is required, because that is what makes this auditable impersonation instead of a silent takeover. It is resolved in this order: an explicit `actor` you pass wins, otherwise the agent session's ID token is used. If the session ID token has expired, the SDK refreshes it first, because Auth0 rejects an expired actor token.
- If no actor can be resolved, the method throws a `TokenExchangeError` with code `actor_unavailable` `before any network call`.
- The STT is `opaque`, single use, and lives about 60 seconds. The SDK never decodes it, never caches it, and never writes it to the session store. The exchange is stateless.
- `buildSessionTransferRedirect` only accepts an absolute `https` URL (plain `http` is allowed for `localhost` only). This stops the single use token from leaking to an untrusted host.
- The two server side error codes (`setactor_required` and `session_transfer_disabled`) are surfaced through the error's raw `cause`, so you get Auth0's exact message.

## Scope

- All changes live `in auth0-server-js only`. There are no changes to `auth0-auth-js` and no dependency bump. This ships as a single, additive minor release.
- The target side needs no new SDK code. You forward `session_transfer_token` (and `organization` when present) through the existing `startInteractiveLogin` call, and read `act` off the session user with `getSession`.

## Prerequisites

These are tenant settings configured outside the SDK. The feature flag `cte_session_transfer_token` must be on. The issuing client needs `can_create_session_transfer_token`,  and the redeeming client needs `allow_delegated_access` with `query` in its allowed authentication methods.

## Testing

- Added tests covering actor sourcing from the session, explicit actor override, ID token refresh on expiry, the `actor_unavailable` client side failure, the multi custom domain audience, the stateless exchange, redirect URL encoding and hardening, and server error surfacing.
- `auth0-server-js`: 381 tests pass. `auth0-auth-js`: 356 tests pass and stays unchanged. Build and lint are clean on both.
- Added an "Impersonation via Session Transfer" guide to `EXAMPLES.md`.

## Summary by CodeRabbit

* **New Features**
  * Added Session Transfer Token (STT) support, including server-side token requests and session-transfer redirect URL generation with organization support and strict validation.
  * Added actor-resolution logic with handling for expired actor ID tokens via refresh, and STT-specific error-code mapping.
  * Exposed STT token-exchange error codes directly from the package entry point.

* **Documentation**
  * Expanded examples with an “Impersonation via Session Transfer” end-to-end flow, including expected `act`-claim usage.

* **Tests**
  * Added comprehensive STT test coverage for validation, parsing, redirect construction, actor/refresh scenarios, and failure mapping.
  * Enhanced token test utilities with configurable expiration.

### DIFF
--- a/packages/auth0-server-js/src/server-client.spec.ts
+++ b/packages/auth0-server-js/src/server-client.spec.ts
@@ -7494,6 +7494,74 @@ test('requestSessionTransferToken - throws ACTOR_UNAVAILABLE when session ID tok
   }
 });
 
+test('requestSessionTransferToken - throws ACTOR_UNAVAILABLE when the session belongs to a different domain (resolver mode)', async () => {
+  const agentIdToken = await generateToken('other.custom.example.com', 'agent_123', '<client_id>');
+  const exchangeSpy = vi.spyOn(AuthClient.prototype, 'exchangeToken');
+
+  const serverClient = new ServerClient({
+    // Resolver mode: the request resolves to a different domain than the session's.
+    domain: async () => 'tenant.custom.example.com',
+    clientId: '<client_id>',
+    clientSecret: '<client_secret>',
+    transactionStore: { get: vi.fn(), set: vi.fn(), delete: vi.fn() },
+    stateStore: {
+      get: vi
+        .fn()
+        .mockResolvedValue({ ...sessionStateWith(agentIdToken, '<refresh_token>'), domain: 'other.custom.example.com' }),
+      set: vi.fn(),
+      delete: vi.fn(),
+      deleteByLogoutToken: vi.fn(),
+    },
+  });
+
+  try {
+    await expect(
+      serverClient.requestSessionTransferToken({
+        subjectToken: 'customer-proof-token',
+        subjectTokenType: 'urn:acme:customer-subject',
+      })
+    ).rejects.toThrowError(expect.objectContaining({ code: 'actor_unavailable' }));
+    expect(exchangeSpy).not.toHaveBeenCalled();
+  } finally {
+    exchangeSpy.mockRestore();
+  }
+});
+
+test('requestSessionTransferToken - throws SessionExpiredError and clears the session when the session_expiry ceiling is reached', async () => {
+  const agentIdToken = await generateToken(domain, 'agent_123', '<client_id>');
+  const exchangeSpy = vi.spyOn(AuthClient.prototype, 'exchangeToken');
+  const mockStateStore = {
+    get: vi
+      .fn()
+      .mockResolvedValue({ ...sessionStateWith(agentIdToken, '<refresh_token>'), sessionExpiresAt: Math.floor(Date.now() / 1000) - 60 }),
+    set: vi.fn(),
+    delete: vi.fn(),
+    deleteByLogoutToken: vi.fn(),
+  };
+
+  const serverClient = new ServerClient({
+    domain,
+    clientId: '<client_id>',
+    clientSecret: '<client_secret>',
+    transactionStore: { get: vi.fn(), set: vi.fn(), delete: vi.fn() },
+    stateStore: mockStateStore,
+  });
+
+  try {
+    await expect(
+      serverClient.requestSessionTransferToken({
+        subjectToken: 'customer-proof-token',
+        subjectTokenType: 'urn:acme:customer-subject',
+      })
+    ).rejects.toThrowError(SessionExpiredError);
+    // The expired session must be cleared, and no exchange attempted.
+    expect(mockStateStore.delete).toHaveBeenCalled();
+    expect(exchangeSpy).not.toHaveBeenCalled();
+  } finally {
+    exchangeSpy.mockRestore();
+  }
+});
+
 test('requestSessionTransferToken - refreshes an expired session ID token and uses the fresh one as actor', async () => {
   const expiredIdToken = await generateToken(domain, 'agent_123', '<client_id>', undefined, undefined, '-1h');
   const freshIdToken = await generateToken(domain, 'agent_123', '<client_id>');
--- a/packages/auth0-server-js/src/server-client.spec.ts
+++ b/packages/auth0-server-js/src/server-client.spec.ts
@@ -2,6 +2,7 @@ import { expect, test, afterAll, afterEach, beforeAll, beforeEach, vi, describe 
 import { ServerClient } from './server-client.js';
 import {
   InvalidConfigurationError,
+  MissingRequiredArgumentError,
   MissingSessionError,
   MissingTransactionError,
   SessionExpiredError,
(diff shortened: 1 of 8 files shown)


packages/auth0-server-js/src/server-client.spec.ts
```diff
@@ -7232,3 +7256,818 @@ describe('logout revocation', () => {
...
+      serverClient.requestSessionTransferToken({ subjectToken: 'customer-proof-token', subjectTokenType: '   ' })
+    ).rejects.toThrowError(MissingRequiredArgumentError);
+    expect(mockStateStore.get).not.toHaveBeenCalled();
+    expect(exchangeSpy).not.toHaveBeenCalled();
+  } finally {
+    exchangeSpy.mockRestore();
+  }
+});
+
+test('requestSessionTransferToken - parses the STT result and branches on issuedTokenType', async () => {
```
kishore7snehil, line 7334: When the server doesn't send back an STT type, the code passes through whatever it got instead of pretending it's an STT, and the docs tell callers to check `issuedTokenType` themselves. There's no test for that case though.

Shall we add one so a non-STT type comes back unchanged and stays that way?

packages/auth0-server-js/src/server-client.spec.ts
```diff
@@ -7232,3 +7256,818 @@ describe('logout revocation', () => {
...
+      })
+    ).rejects.toThrowError(expect.objectContaining({ code: 'actor_unavailable' }));
+    expect(exchangeSpy).not.toHaveBeenCalled();
+  } finally {
+    refreshSpy.mockRestore();
+    exchangeSpy.mockRestore();
+  }
+});
+
+test('requestSessionTransferToken - surfaces SETACTOR_REQUIRED as a TokenExchangeError from the server', async () => {
```
kishore7snehil, line 7820: This tests `setactor_required` and checks the server's message comes through on `cause`. The other server error, `session_transfer_disabled`, takes the exact same path but has no test.

Worth adding a matching one so both server errors are covered.

packages/auth0-server-js/src/server-client.ts
```diff
@@ -1246,6 +1282,228 @@ export class ServerClient<TStoreOptions = unknown> {
...
+    // belongs to a different domain than the one resolved for this request.
+    if (this.#isResolverMode() && !(await this.#isSessionForCurrentDomain(stateData, storeOptions))) {
+      throw actorUnavailableError(
+        'Unable to resolve an actor for the session transfer token: the agent session belongs to a different domain than the one resolved for this request. Pass an explicit actor or ensure the agent is logged in on this domain.'
+      );
+    }
+
+    // A session past its `session_expiry` ceiling must not keep minting tokens; clear it and fail,
+    // mirroring getAccessToken / startLinkUser.
+    if (isSessionExpiryReached(stateData.sessionExpiresAt)) {
```
kishore7snehil: This blocks an STT when the session has passed its `session_expiry` time. But a session only has that time set if the agent logged in through an IdP that sends `session_expiry`, and there's no guarantee an agent session has it. When it's missing, this check does nothing.

packages/auth0-server-js/src/server-client.spec.ts
```diff
@@ -7232,3 +7256,818 @@ describe('logout revocation', () => {
...
+        subjectTokenType: 'urn:acme:customer-subject',
+      })
+    ).rejects.toThrowError(expect.objectContaining({ code: 'actor_unavailable' }));
+    expect(exchangeSpy).not.toHaveBeenCalled();
+  } finally {
+    exchangeSpy.mockRestore();
+  }
+});
+
+test('requestSessionTransferToken - throws ACTOR_UNAVAILABLE when the session belongs to a different domain (resolver mode)', async () => {
```
kishore7snehil, line 7632: This tests the resolver case where the session's domain doesn't match. The opposite case, where they do match and the exchange runs, isn't tested. The other domain test mocks `exchangeToken`, so it skips the real session read.

Shall we add one so both sides are covered?

packages/auth0-server-js/src/server-client.ts
```diff
@@ -72,6 +78,36 @@ const decodeIssuer = (token: string) => {
...
+// A small skew so an ID token about to expire is treated as expired and refreshed, rather than
+// being sent as an actor token and rejected by the server on its `exp` check.
+const ID_TOKEN_EXPIRY_SKEW_SECONDS = 30;
+
+// Builds the client-side `actor_unavailable` failure for the Session Transfer Token flow. Reuses
+// the core `TokenExchangeError` (the same error the exchange surfaces) and stamps the STT-specific
+// code, so callers get a consistent error surface without a bespoke error class.
+const actorUnavailableError = (message: string): TokenExchangeError => {
+  const error = new TokenExchangeError(message);
+  error.code = TokenExchangeErrorCode.ACTOR_UNAVAILABLE;
```
kishore7snehil, line 93: This sets `code` on the error after creating it because `TokenExchangeError` always starts with `token_exchange_error` and has no way to pass the code in. So the error briefly carries the wrong code before this line fixes it to `actor_unavailable`.

kishore7snehil (changes requested):

packages/auth0-server-js/src/server-client.ts
nandan-bhat: The only clean way to pass the `code` in at construction time would be to change the  `TokenExchangeError` constructor itself. That class lives in the core `auth0-auth-js` package and is part of its public API. Its constructor hardcodes `token_exchange_error` as the code, the same way every other error class in that file hardcodes its own. Adding a code parameter there is a wider change than it looks, and we already have a backward compatibility note on a sibling class (`TokenForConnectionError`) about consumers branching on these values, so I would rather not touch that surface just for this.

On the `"briefly carries the wrong code"` point, the reassignment is fully synchronous and local to the helper. Nothing reads the error between the two lines, and the object is only handed to the caller after the code has been set to `actor_unavailable`. So there is no window where anything can observe the intermediate value. It reads a little awkward, but it is correct.

The helper exists precisely so this reuse stays in one place and callers get a consistent error shape without us adding a whole new error class. If we later decide the core error classes should accept a code, I am happy to move this over as part of that change. For this PR I would prefer to leave it as is. Let me know if you feel strongly and we can revisit.

packages/auth0-server-js/src/server-client.spec.ts
nandan-bhat: Addressed in 28d82f3ec60fd2863d6d7a7a73b0e5dd6457b9be

packages/auth0-server-js/src/server-client.spec.ts
nandan-bhat: Addressed in 28d82f3ec60fd2863d6d7a7a73b0e5dd6457b9be

packages/auth0-server-js/src/server-client.ts
nandan-bhat: Addressed in 28d82f3ec60fd2863d6d7a7a73b0e5dd6457b9be

packages/auth0-server-js/src/server-client.spec.ts
nandan-bhat: Addressed in 28d82f3ec60fd2863d6d7a7a73b0e5dd6457b9be

kishore7snehil (approved):
LGTM!

PR #222, opened 2026-07-15
feat(auth0-server-js): add revokeRefreshToken and logout revocation support

jd3vi1:
## Summary

- Adds `ServerClient.revokeRefreshToken()` — reads the refresh token from the session and delegates to `AuthClient.revokeToken()`, wrapping errors in `TokenRevocationError`
- `logout()` now always attempts revocation best-effort before clearing the session; revocation failure never blocks logout
- In resolver mode, both revocation and session deletion are gated by a domain-match guard (session domain must match resolved domain)
- Re-exports `TokenRevocationError` from the server-js public surface
- Adds `/auth/revoke` route to the Express example app
- Documents revocation behavior in EXAMPLES.md

## Depends on

Requires auth0/auth0-auth-js#221 to be merged and released first.

## Test plan

- [x] `revokeRefreshToken` — success, missing session, missing refresh token, wraps error as `TokenRevocationError`
- [x] `logout` (static mode) — revokes then deletes session; revocation failure does not block logout
- [x] `logout` (resolver mode) — revokes and deletes when domains match; skips both when domains differ

### DIFF
--- a/examples/example-express-web/src/auth0.ts
+++ b/examples/example-express-web/src/auth0.ts
@@ -84,5 +84,10 @@ export function auth0(options: Auth0ExpressOptions) {
     response.redirect(logoutUrl.href);
   });
 
+  router.post('/auth/revoke', async (request: Request, response: Response) => {
+    await request.auth0Client.revokeRefreshToken({}, { request, response });
+    response.redirect('/private');
+  });
+
   return router;
 }
--- a/examples/example-express-web/src/auth0.ts
+++ b/examples/example-express-web/src/auth0.ts
@@ -32,6 +32,7 @@ export function auth0(options: Auth0ExpressOptions) {
     clientSecret: options.clientSecret,
     authorizationParams: {
       redirect_uri: redirectUri.toString(),
+      scope: 'openid profile email offline_access',
     },
     transactionStore: new CookieTransactionStore(
       {
--- a/packages/auth0-server-js/src/index.ts
+++ b/packages/auth0-server-js/src/index.ts
@@ -2,7 +2,7 @@ export { ServerClient } from './server-client.js';
 export { AbstractStateStore } from './store/abstract-state-store.js';
 export { AbstractTransactionStore } from './store/abstract-transaction-store.js';
 export type { TokenResponse, ActClaim } from '@auth0/auth0-auth-js';
-export { TokenExchangeError, MissingClientAuthError, OrganizationValidationError } from '@auth0/auth0-auth-js';
+export { TokenExchangeError, TokenRevocationError, MissingClientAuthError, OrganizationValidationError } from '@auth0/auth0-auth-js';
 
 export type { CookieHandler, CookieSerializeOptions } from './store/cookie-handler.js';
 export { CookieTransactionStore } from './store/cookie-transaction-store.js';
--- a/packages/auth0-server-js/src/server-client.spec.ts
+++ b/packages/auth0-server-js/src/server-client.spec.ts
@@ -6924,38 +6924,65 @@ describe('revokeRefreshToken', () => {
     await expect(serverClient.revokeRefreshToken()).rejects.toThrowError(TokenRevocationError);
   });
 
-  test('should use session domain in resolver mode', async () => {
-    // Resolver returns a different domain than the one stored in the session.
-    // The revoke handler is only registered on the session domain (session.local)
-    // so the test will fail with a network error if the resolver domain
-    // (resolver.local) is used instead - proving session domain takes precedence.
-    const domainResolver = vi.fn().mockResolvedValue('resolver.local');
+  test('should revoke in resolver mode when session domain matches resolved domain', async () => {
+    const domainResolver = vi.fn().mockResolvedValue('matched.local');
 
-    let sessionDomainRevokeCalled = false;
+    let revokeCalled = false;
 
     server.use(
-      http.get('https://session.local/.well-known/openid-configuration', () => {
-        return HttpResponse.json({
-          issuer: 'https://session.local/',
-          authorization_endpoint: 'https://session.local/authorize',
-          token_endpoint: 'https://session.local/token',
-          end_session_endpoint: 'https://session.local/logout',
-          revocation_endpoint: 'https://session.local/oauth/revoke',
-          jwks_uri: 'https://session.local/.well-known/jwks.json',
-        });
-      }),
-      http.post('https://session.local/oauth/revoke', () => {
-        sessionDomainRevokeCalled = true;
+      http.get('https://matched.local/.well-known/openid-configuration', () =>
+        HttpResponse.json({
+          issuer: 'https://matched.local/',
+          authorization_endpoint: 'https://matched.local/authorize',
+          token_endpoint: 'https://matched.local/token',
+          end_session_endpoint: 'https://matched.local/logout',
+          revocation_endpoint: 'https://matched.local/oauth/revoke',
+          jwks_uri: 'https://matched.local/.well-known/jwks.json',
+        })
+      ),
+      http.post('https://matched.local/oauth/revoke', () => {
+        revokeCalled = true;
         return new HttpResponse(null, { status: 200 });
       })
     );
 
-    const mockStateStore = {
-      get: vi.fn(),
-      set: vi.fn(),
-      delete: vi.fn(),
-      deleteByLogoutToken: vi.fn(),
+    const stateData: StateData = {
+      user: { sub: '<sub>' },
+      idToken: '<id_token>',
+      refreshToken: '<refresh_token>',
+      tokenSets: [],
+      domain: 'matched.local',
+      internal: { sid: '<sid>', createdAt: Date.now() },
     };
+
+    const serverClient = new ServerClient({
+      domain: domainResolver,
+      clientId: '<client_id>',
+      clientSecret: '<client_secret>',
+      discoveryCache: { ttl: 0 },
+      transactionStore: { get: vi.fn(), set: vi.fn(), delete: vi.fn() },
+      stateStore: { get: vi.fn().mockResolvedValue(stateData), set: vi.fn(), delete: vi.fn(), deleteByLogoutToken: vi.fn() },
+    });
+
+    await expect(serverClient.revokeRefreshToken()).resolves.toBeUndefined();
+    expect(revokeCalled).toBe(true);
+  });
+
+  test('should skip revocation in resolver mode when session domain does not match resolved domain', async () => {
+    const domainResolver = vi.fn().mockResolvedValue('resolver.local');
+
+    let revokeCalled = false;
+
+    server.use(
+      http.post('https://session.local/oauth/revoke', () => {
+        revokeCalled = true;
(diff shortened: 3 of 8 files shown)


packages/auth0-server-js/src/server-client.ts
```diff
@@ -1077,6 +1078,53 @@ export class ServerClient<TStoreOptions = unknown> {
...
+    }
+
+    if (!refreshToken) {
+      throw new MissingSessionError('Unable to revoke refresh token: no refresh token found in session.');
+    }
+
+    let authClient: AuthClient;
+    if (this.#isResolverMode()) {
+      const resolvedDomain = await this.#resolveDomain(storeOptions);
+      const sessionDomain = stateData ? this.#getSessionDomain(stateData) : undefined;
```
tusharpandey13, line 1125: code is missing the resolvedDomain guard, see https://github.com/auth0/auth0-auth-js/blob/be49b125b2544c82f540170fae7d43a3a75b278f/packages/auth0-server-js/src/server-client.ts#L1102

`sessionDomain ?? #resolveDomain()` sends a session refresh token to the current-request tenant when session has no stored domain. Wrong-tenant revoke → origin token survives. Use `strict sessionDomain === resolvedDomain` guard

packages/auth0-server-js/src/types.ts
tusharpandey13, line 1: Can revocation be made opt-in using a config param?

packages/auth0-server-js/src/server-client.ts
```diff
@@ -1085,21 +1133,35 @@ export class ServerClient<TStoreOptions = unknown> {
    */
   public async logout(options: LogoutOptions, storeOptions?: TStoreOptions) {
     if (!this.#isResolverMode()) {
+      try {
+        await this.revokeRefreshToken({}, storeOptions);
+      } catch (e) {
+        // best-effort: revocation failure must not block logout
+        console.warn('revokeRefreshToken failed during logout (swallowed):', e);
```
tusharpandey13: code smell: console log

packages/auth0-server-js/src/server-client.spec.ts
```diff
@@ -6774,3 +6776,426 @@ test('changePassword resolves the domain in resolver mode (T4.3)', async () => {
...
+  const setupRevocation = (handler?: Parameters<typeof http.post>[1]) => {
+    server.use(
+      http.get(`https://${domain}/.well-known/openid-configuration`, () =>
+        HttpResponse.json({ ...mockOpenIdConfiguration, revocation_endpoint: revocationEndpoint })
+      ),
+      http.post(revocationEndpoint, handler ?? (() => new HttpResponse(null, { status: 200 })))
+    );
+  };
+
+  test('should revoke the refresh token from the session', async () => {
```
tusharpandey13, line 6791: only asserts resolves.toBeUndefined()

packages/auth0-server-js/src/server-client.spec.ts
```diff
@@ -6774,3 +6776,426 @@ test('changePassword resolves the domain in resolver mode (T4.3)', async () => {
...
+      discoveryCache: { ttl: 0 },
+      transactionStore: { get: vi.fn(), set: vi.fn(), delete: vi.fn() },
+      stateStore: { get: vi.fn().mockResolvedValue(stateData), set: vi.fn(), delete: vi.fn(), deleteByLogoutToken: vi.fn() },
+    });
+
+    await expect(serverClient.revokeRefreshToken()).resolves.toBeUndefined();
+    expect(revokeCalled).toBe(true);
+  });
+
+  test('should skip revocation in resolver mode when session domain does not match resolved domain', async () => {
```
tusharpandey13, line 6969: no negative assertion that revoke was NOT called

packages/auth0-server-js/src/server-client.ts
```diff
@@ -1077,6 +1078,53 @@ export class ServerClient<TStoreOptions = unknown> {
...
+      if (sessionDomain && sessionDomain !== resolvedDomain) {
+        // Session belongs to a different tenant; do not revoke against the wrong domain.
+        return;
+      }
+      authClient = this.#getAuthClient(sessionDomain ?? resolvedDomain);
+    } else {
+      authClient = this.authClient;
+    }
+
+    await authClient.revokeToken({ token: refreshToken, tokenTypeHint: 'refresh_token' });
```
tusharpandey13, line 1135: `option.token` accepts any token. is hardcoding okay here?

tusharpandey13 (commented):
PTAL

packages/auth0-server-js/src/types.ts
jd3vi1: I'd kept it as opt-in initially, but after discussing w/ Piyush we made it default, since similar thing was flagged in nextjs for a security issue.

packages/auth0-server-js/src/server-client.ts
jd3vi1: this func is specifically for revoking refresh token, so kept it hardcoded.

examples/example-express-web/src/auth0.ts
```diff
@@ -84,5 +84,10 @@ export function auth0(options: Auth0ExpressOptions) {
     response.redirect(logoutUrl.href);
   });
 
+  router.post('/auth/revoke', async (request: Request, response: Response) => {
+    await request.auth0Client.revokeRefreshToken({}, { request, response });
+    response.redirect('/private');
+  });
```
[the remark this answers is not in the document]
jd3vi1: Fixed - added `scope: 'openid profile email offline_access'` to `authorizationParams` in the example.

packages/auth0-server-js/src/index.ts
```diff
@@ -2,7 +2,7 @@ export { ServerClient } from './server-client.js';
 export { AbstractStateStore } from './store/abstract-state-store.js';
 export { AbstractTransactionStore } from './store/abstract-transaction-store.js';
 export type { TokenResponse, ActClaim } from '@auth0/auth0-auth-js';
-export { TokenExchangeError, MissingClientAuthError, OrganizationValidationError } from '@auth0/auth0-auth-js';
+export { TokenExchangeError, TokenRevocationError, MissingClientAuthError, OrganizationValidationError } from '@auth0/auth0-auth-js';
```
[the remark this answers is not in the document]
jd3vi1: Fixed - bumped the `@auth0/auth0-auth-js` dependency to `^1.12.0` which includes `TokenRevocationError`. The re-export is now valid.

packages/auth0-server-js/src/server-client.ts
```diff
@@ -1077,6 +1078,53 @@ export class ServerClient<TStoreOptions = unknown> {
     };
   }
 
+  /**
+   * Revokes the refresh token stored in the current session, or an explicitly supplied token.
+   *
+   * @param options Optionally supply a token to revoke instead of reading from the session.
+   * @param storeOptions Optional options passed to the StateStore.
```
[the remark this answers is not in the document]
jd3vi1: Fixed - updated the JSDoc to explicitly state that the domain-match guard applies in resolver mode even when a token is passed explicitly. Also updated the corresponding section in EXAMPLES.md.

packages/auth0-server-js/src/server-client.ts
```diff
@@ -1077,6 +1078,53 @@ export class ServerClient<TStoreOptions = unknown> {
...
+    const stateData = needsStateData
+      ? await this.#stateStore.get(this.#stateStoreIdentifier, storeOptions)
+      : undefined;
+
+    if (!refreshToken) {
+      refreshToken = stateData?.refreshToken;
+    }
+
+    if (!refreshToken) {
+      throw new MissingSessionError('Unable to revoke refresh token: no refresh token found in session.');
```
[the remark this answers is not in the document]
jd3vi1: Fixed - added an explicit guard that throws `MissingRequiredArgumentError` when `options.token` is an empty string, before the session fallback logic.

packages/auth0-server-js/src/server-client.ts
```diff
@@ -1077,6 +1078,53 @@ export class ServerClient<TStoreOptions = unknown> {
...
+
+    let authClient: AuthClient;
+    if (this.#isResolverMode()) {
+      const resolvedDomain = await this.#resolveDomain(storeOptions);
+      const sessionDomain = stateData ? this.#getSessionDomain(stateData) : undefined;
+      if (sessionDomain && sessionDomain !== resolvedDomain) {
+        // Session belongs to a different tenant; do not revoke against the wrong domain.
+        return;
+      }
+      authClient = this.#getAuthClient(sessionDomain ?? resolvedDomain);
```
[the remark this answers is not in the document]
jd3vi1: Fixed in a prior commit - the guard was updated to `if (stateData && sessionDomain !== resolvedDomain)`. This treats any session with an undetermined domain as a mismatch and returns without revoking, rather than falling back to the resolved tenant.

packages/auth0-server-js/src/server-client.ts
```diff
@@ -1085,21 +1133,35 @@ export class ServerClient<TStoreOptions = unknown> {
    */
   public async logout(options: LogoutOptions, storeOptions?: TStoreOptions) {
     if (!this.#isResolverMode()) {
+      try {
+        await this.revokeRefreshToken({}, storeOptions);
+      } catch (e) {
+        // best-effort: revocation failure must not block logout
+        console.warn('revokeRefreshToken failed during logout (swallowed):', e);
+      }
```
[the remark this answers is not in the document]
jd3vi1: Both logout catch blocks are already silent (no `console.warn`) - all revocation errors including `MissingSessionError` are swallowed silently so logout always proceeds cleanly.

packages/auth0-server-js/src/server-client.ts
jd3vi1: Fixed - the guard is now `if (stateData && sessionDomain !== resolvedDomain)`. Sessions with no stored domain are treated as a mismatch, so they never reach the revocation call against the wrong tenant. A dedicated test covers this case.

packages/auth0-server-js/src/server-client.ts
jd3vi1: Fixed - removed `console.warn` from both logout catch blocks. Both are now bare `catch {}` with only a comment.

packages/auth0-server-js/src/server-client.spec.ts
jd3vi1: Fixed - the test now captures the token sent to the revoke endpoint and asserts it matches the expected value, not just that the call resolved.

packages/auth0-server-js/src/server-client.spec.ts
jd3vi1: Fixed - the resolver mode mismatch tests now assert `expect(revokeCalled).toBe(false)` to explicitly confirm revocation was skipped.

packages/auth0-server-js/src/index.ts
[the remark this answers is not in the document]
jd3vi1: #226 is done and released the newer version

tusharpandey13 (approved):

PR #228, opened 2026-07-17
feat: add getUserInfo() to auth0-auth-js

tusharpandey13:
### Description

Adds `getUserInfo()` to the `AuthClient` for retrieving a user's OIDC profile claims from the `/userinfo` endpoint, and exposes a thin delegating `getUserInfo()` on `ServerClient`.

```ts
const userInfo = await authClient.getUserInfo({
  accessToken: '<access_token>',
});

console.log(userInfo.sub, userInfo.email, userInfo.name);
```

Pass an optional `expectedSubject` to assert the returned `sub` matches a known value; a mismatch throws `UserInfoError`.

### What's included

- `AuthClient.getUserInfo(options, requestOptions?)` accepting `{ accessToken, expectedSubject? }` and an optional trailing `RequestOptions` (`signal`, `headers`, `customFetch`), consistent with the other network-performing methods.
- `ServerClient.getUserInfo(options, storeOptions?, requestOptions?)` delegating to auth-js and resolver-domain aware. Types and `UserInfoError` are re-exported from `@auth0/auth0-server-js` so consumers do not reach into `@auth0/auth0-auth-js`.
- `UserInfoError`, thrown when the `/userinfo` request fails or a subject check does not match. Carries `statusCode`/`headers` when the response provides them (`set-cookie` is stripped).
- `GetUserInfoOptions` and `UserInfoResponse` types. `UserInfoResponse` includes an index signature so additional claims are forward-compatible.

`/userinfo` is bearer-protected and requires no client authentication, so `getUserInfo()` works for public clients (client id only): the supplied access token is the only credential used.

### Access token requirements

The access token must be one that `/userinfo` accepts:

- **Without Multi-Resource Refresh Tokens (MRRT):** use a default OIDC access token, obtained without an explicit `audience`.
- **With MRRT:** access tokens are audience-bound, so request the userinfo endpoint as the audience (for example, `https://<your-domain>/userinfo`) when obtaining the token. A token bound to a different audience is rejected by `/userinfo`.

See `EXAMPLES.md` for full usage, including subject validation and error handling.

## Summary by CodeRabbit

* **New Features**
  * Added UserInfo retrieval through the OIDC `/userinfo` endpoint for authentication and server clients.
  * Added optional subject validation and UserInfo-specific error details, including HTTP status and request identifiers.
  * Added per-request network controls, including custom headers, abort signals, and custom fetch implementations.
  * Added optional full HTTP responses and expanded response metadata.
  * Added support for UserInfo requests from public clients and forwarding request options through server clients.

* **Documentation**
  * Added UserInfo examples covering token requirements, returned claims, validation, and error handling.


tusharpandey13:
@coderabbitai review

Piyush-85 (approved):
LGTM. MRRT docs are correct, types exported via wildcards, tests comprehensive (13 cases). One cosmetic nit: PR description is stale (still says 'no server-js getUserInfo' and old MRRT wording) -- worth updating for posterity but not blocking.

Piyush-85 (dismissed):
LGTM !

packages/auth0-auth-js/src/auth-client.ts
```diff
@@ -1670,6 +1673,59 @@ export class AuthClient {
...
+   * @example With subject validation
+   * ```typescript
+   * const userInfo = await authClient.getUserInfo({
+   *   accessToken: myAccessToken,
+   *   expectedSubject: knownSubjectId
+   * });
+   * ```
+   */
+  public async getUserInfo(options: GetUserInfoOptions): Promise<UserInfoResponse> {
+    const { configuration } = await this.#discover();
```
nandan-bhat: `#discover()` goes through `#getClientAuth()`, which throws `MissingClientAuthError` when there is no `clientSecret`, `clientAssertionSigningKey` or `useMtls`. I tried this with a public client and got `MissingClientAuthError` back, not `UserInfoError`.

`/userinfo` does not need client authentication at all. The only reason we touch discovery here is to read `userinfo_endpoint`. So a public client that already holds a valid access token has no way to call this method, and the error it gets does not match the `@throws {UserInfoError}` on the JSDoc above.

The README says public clients are supported for `register()` and `challenge()`, so this reads like an accidental restriction rather than a decision. Can we resolve the metadata without requiring client auth on this path? If that is too much for this PR, could we at least document the confidential client requirement and add a test, so the behaviour is deliberate?

packages/auth0-auth-js/src/auth-client.ts
```diff
@@ -1670,6 +1673,59 @@ export class AuthClient {
...
+    try {
+      return await client.fetchUserInfo(
+        configuration,
+        options.accessToken,
+        options.expectedSubject ?? client.skipSubjectCheck
+      );
+    } catch (e) {
+      throw new UserInfoError(
```
nandan-bhat, line 1771: `toOAuth2Error` only looks at `err.status` and `err.response`. When `/userinfo` returns a non 200 and the response has no `WWW-Authenticate` header, `openid-client` throws an `OperationProcessingError`, which has neither of those, so we drop all of it.

What I got when I ran this:

* 401 with `WWW-Authenticate`: `statusCode` is `401`, `headers` present, `cause.error` is `''`
* 401 without `WWW-Authenticate`: `statusCode` is `undefined`, `headers` is `undefined`, `cause.error` is `''`
* 429 with `Retry-After: 30`: `statusCode` is `undefined`, `headers` is `undefined`, `cause.error` is `''`

The data is still reachable though. On that error `cause` is the `Response` itself, so `err.cause.status` is `429` and `err.cause.headers.get('retry-after')` is `30`.

This one matters because `examples/http-requests-and-responses.md` says `statusCode` and `headers` are available on all error types, and shows a `429` plus `Retry-After` backoff recipe as the recommended pattern. `/userinfo` is rate limited, so anyone following that recipe here gets a `429` branch that never runs and no `x-request-id` to give support.

Could `toOAuth2Error` also read from `cause` when `cause instanceof Response`? It is a small change and it would help every other path that lands on this branch. `UserInfoError` also needs adding to the error list in that docs page.

packages/auth0-auth-js/src/auth-client.ts
```diff
@@ -1670,6 +1673,59 @@ export class AuthClient {
...
+   *
+   * @example With subject validation
+   * ```typescript
+   * const userInfo = await authClient.getUserInfo({
+   *   accessToken: myAccessToken,
+   *   expectedSubject: knownSubjectId
+   * });
+   * ```
+   */
+  public async getUserInfo(options: GetUserInfoOptions): Promise<UserInfoResponse> {
```
nandan-bhat: This makes a real HTTP call but takes no `RequestOptions`, so there is no way to pass a `signal`, per request `headers`, or a one off `customFetch`.

`examples/http-requests-and-responses.md` says every network performing method on `AuthClient` accepts a trailing `RequestOptions`, and the only carve out it lists is the four URL builders. Right now a reader who follows that doc will write `getUserInfo(opts, { signal })` and the second argument gets quietly ignored.

Using `#discoverForRequest(requestOptions)` instead of `#discover()` should cover it, same as `getTokenByRefreshToken`.

packages/auth0-auth-js/src/types.ts
```diff
@@ -913,3 +913,65 @@ export interface ApiResponse<T> {
...
 }
+
+/**
+ * Options for retrieving user information from the /userinfo endpoint.
+ */
+export interface GetUserInfoOptions {
+  /** The access token to send to the /userinfo endpoint. */
+  accessToken: string;
+  /**
+   * Expected `sub` claim. When provided, the returned subject is validated
```
nandan-bhat, line 924: Not asking for a change, I just want the choice visible in review.

The default is `skipSubjectCheck`, and `openid-client` flags that one with a warning about security implications. The reason given here is `node-auth0` parity, but parity with a different SDK is not really a compatibility constraint on a brand new API in this one.

That said, the risk is low in practice because the caller supplies the token, and the advisory note here plus the `EXAMPLES` callout are a fair landing point. Fine by me to ship as is.

packages/auth0-auth-js/examples/user-info.md
```diff
@@ -0,0 +1,76 @@
...
+
+try {
+  const userInfo = await authClient.getUserInfo({
+    accessToken: myAccessToken,
+  });
+} catch (error) {
+  if (error instanceof UserInfoError) {
+    console.error('Failed to retrieve user info:', error.message);
+    console.error('Error code:', error.code);           // 'user_info_error'
+    console.error('OAuth error:', error.cause?.error);  // e.g., 'unauthorized'
```
nandan-bhat: `error.cause?.error` is always an empty string here, it is never `'unauthorized'`. I checked 401 with and without `WWW-Authenticate`, 403, 429 and a malformed body, and got `''` every time. `cause.error_description` is empty too.

Auth0 does send an error body, but `openid-client` throws before anything parses it, so we never see it.

Once the metadata question on `auth-client.ts` is settled, this snippet reads better around `error.statusCode` and `error.headers?.get('x-request-id')`. The `error.code` line just above is correct.

packages/auth0-auth-js/src/auth-client.spec.ts
```diff
@@ -5074,3 +5121,205 @@ describe('negative / fullResponse edge cases', () => {
...
+describe('getUserInfo', () => {
+  const makeClient = () =>
+    new AuthClient({
+      domain,
+      clientId: '<client_id>',
+      clientSecret: '<client_secret>',
+    });
+
+  afterEach(() => {
+    // After outer afterEach resets handlers, restore restHandlers for next test
```
nandan-bhat: I think this block is a no op. Vitest runs the inner `afterEach` before the outer one, so the outer `server.resetHandlers()` runs after this and restores the original handlers anyway. I deleted the whole block locally and all 221 tests still passed.

The comment also has the ordering the other way round, which makes it a bit misleading to read. Can we drop it?

packages/auth0-auth-js/src/auth-client.spec.ts
```diff
@@ -5074,3 +5121,205 @@ describe('negative / fullResponse edge cases', () => {
...
+    const client = makeClient();
+
+    // When no expectedSubject is provided, openid-client.skipSubjectCheck is used
+    const result = await client.getUserInfo({ accessToken });
+
+    expect(result.sub).toBe('user_123');
+    expect(result.email).toBe('jane@example.com');
+  });
+
+  test('A5 - HTTP 401 Unauthorized', async () => {
```
nandan-bhat, line 5159: The `/userinfo` handlers return error bodies with no `WWW-Authenticate` header, and that sends us down the `OperationProcessingError` path. A response that does carry the header goes through `WWWAuthenticateChallengeError` instead and gives a different error object, `statusCode` is `undefined` in one case and `401` in the other.

Auth0 serves `/userinfo` as a normal bearer protected resource, so the header present path is most likely the one users actually hit. A5, A6 and A12 only assert `name` and `code`, and those are identical on both paths, so neither shape is really pinned.

Could we add a handler variant that sets `WWW-Authenticate: Bearer realm="...", error="invalid_token"` and assert `statusCode` on both?

packages/auth0-auth-js/src/auth-client.spec.ts
```diff
@@ -5074,3 +5121,205 @@ describe('negative / fullResponse edge cases', () => {
...
+    });
+
+    const err = await client.getUserInfo({ accessToken }).catch((e) => e);
+
+    expect(err).toBeInstanceOf(Error);
+    expect(err.name).toBe('UserInfoError');
+    expect(err.code).toBe('user_info_error');
+  });
+
+  test('A10 - Calls discovery for metadata', async () => {
```
nandan-bhat: A1, A2, A4 and A10 all call `getUserInfo({ accessToken })` against the same default handler and assert `sub`. The comment here says it only checks that the method succeeds, so the name `Calls discovery for metadata` promises more than the test does. A4 is in the same spot for `skipSubjectCheck`.

Could we fold A2, A4 and A10 into A1? If discovery is worth pinning, counting the `/.well-known/openid-configuration` requests across two `getUserInfo` calls would actually test it.

packages/auth0-auth-js/src/auth-client.spec.ts
```diff
@@ -5074,3 +5121,205 @@ describe('negative / fullResponse edge cases', () => {
...
+        expectedSubject: 'user_correct',
+      })
+      .catch((e) => e);
+
+    expect(err).toBeInstanceOf(Error);
+    expect(err.name).toBe('UserInfoError');
+    expect(err.code).toBe('user_info_error');
+  });
+
+  test('A8b - Regression: empty string expectedSubject is treated as mismatch, not skip', async () => {
```
nandan-bhat: The behaviour being locked down is right, but the reason in the name and comment is not what happens. `expectedSubject: ''` fails inside `oauth4webapi`'s `assertString` with `"expectedSubject" must not be empty`, before any `sub` comparison runs. I confirmed by reading `cause.message` off the thrown error.

Since this comment exists to stop someone flipping `??` back to `||`, it would help if it described the real mechanism. Maybe rename to something like `empty string expectedSubject is rejected, not silently skipped`.

packages/auth0-server-js/src/server-client.ts
```diff
@@ -1038,6 +1038,33 @@ export class ServerClient<TStoreOptions = unknown> {
...
+   *   (e.g. `https://<domain>/userinfo`) when obtaining the token. A token bound to a
+   *   different resource-server audience is rejected by `/userinfo`.
+   *
+   * @param options Options containing the access token and an optional expected subject
+   *                for OIDC subject-consistency validation.
+   * @param storeOptions Optional store options, used to resolve the domain in resolver mode.
+   * @throws {UserInfoError} When the `/userinfo` request fails or the subject check fails.
+   * @returns A Promise resolving to the UserInfo claims.
+   */
+  public async getUserInfo(
```
nandan-bhat, line 1064: Two small things on this signature.

`EXAMPLES.md` keeps a `Methods that do not accept RequestOptions` list with a reason for every entry, and it is exhaustive today. `getUserInfo` is in neither that list nor the supported table, so it is left undefined either way.

If we thread `RequestOptions` through `auth0-auth-js`, this one wants the documented `(options, storeOptions?, requestOptions?)` shape. If we do not, could you add it to the exclusion list with the reason?

The `#resolveDomain` plus `#getAuthClient` bit is correct, I checked resolver mode works here.

packages/auth0-server-js/EXAMPLES.md
```diff
@@ -1638,6 +1689,7 @@ const result = await serverClient.getAccessToken({ ...baseOpts, fullResponse: tr
 
 If the wrong overload is selected at compile time (because `fullResponse` widened to `boolean`), you will see a TypeScript error when trying to access `result.response`, or your code will assume `result` is a `TokenSet` when it should be an `ApiResponse<TokenSet>`.
 
+
```
nandan-bhat: Looks like a stray blank line, there are two in a row before this heading now. Unrelated to the feature.

While you are in this file: the `#passing-storeoptions-N` renumbering matches `main`, but `main` is already off by one from `Retrieving the logged-in User` onward. There are 15 `### Passing StoreOptions` headings and only 14 TOC entries, because the one under `## Passwordless Authentication` has no TOC line. Not caused by this PR, but you are already rewriting these lines so it is close to free to fix.

packages/auth0-server-js/EXAMPLES.md
```diff
@@ -1182,6 +1184,55 @@ const tokenResponse = await serverClient.customTokenExchange({ subjectToken, sub
...
+    expectedSubject: '<known_sub>', // optional; throws UserInfoError on mismatch
+  });
+
+  console.log(userInfo.sub);
+  console.log(userInfo.email);
+  console.log(userInfo.name);
+} catch (error) {
+  if (error instanceof UserInfoError) {
+    console.error('Failed to retrieve user info:', error.message);
+    console.error('OAuth error:', error.cause?.error); // e.g., 'unauthorized'
```
nandan-bhat: Same as the `auth0-auth-js` example, `error.cause?.error` is always `''` here, so this line points people at an empty value.

packages/auth0-server-js/EXAMPLES.md
tusharpandey13: will push unrelated doc fixes in a seperate PR, to keep scope of this PR clean

packages/auth0-auth-js/src/auth-client.ts
tusharpandey13: Fixed in 3f11703. `getUserInfo` now discovers with client auth optional: a new `clientAuthOptional` flag is threaded through `#discover`/`#discoverForRequest`/`#createConfiguration`/`#getClientAuth`, and when set with no credentials it returns `client.None()` instead of throwing. `/userinfo` is bearer-protected and needs no client auth, so a public client with a valid access token now reaches the endpoint and surfaces `UserInfoError` as documented. The `None()` path deliberately bypasses the memoized client-auth promise and the strict instance config cache so it can never poison a later confidential call.

packages/auth0-auth-js/src/auth-client.ts
tusharpandey13: Confirmed and partly addressed in 3f11703. The two error shapes come from the underlying client: responses carrying `WWW-Authenticate` throw a structured error (`status` + `response`), which `toOAuth2Error` already captures; responses without it (429, gateway 5xx, malformed) throw `OperationProcessingError`, which genuinely carries neither `status` nor `response`. Rather than fabricate values, I documented this on the JSDoc: when no `WWW-Authenticate` header is present, `statusCode`/`headers` are left unset. Auth0 `/userinfo` sends the header on 401/403, so the common path populates both; A15 now pins that.

packages/auth0-auth-js/src/auth-client.ts
tusharpandey13: Fixed in 3f11703. `getUserInfo(options, requestOptions?)` now accepts a trailing `RequestOptions` and resolves via `#discoverForRequest(requestOptions, true)`, so `signal`/`headers`/`customFetch` take effect like every other network method. `ServerClient.getUserInfo` forwards it as its third argument. Added tests for signal propagation (auth-js) and forwarding (server-js).

packages/auth0-auth-js/src/types.ts
tusharpandey13: Fair point on the parity framing, agreed it is not a compat constraint on a new API. Keeping the default as-is though, for a different reason: subject validation compares the userinfo `sub` against a caller-supplied `expectedSubject`, so with none supplied there is nothing to validate against and skip is the only coherent default. Making `expectedSubject` required would add migration friction (every call site needs a known sub) with no security gain when the caller already holds the token. The advisory note plus the EXAMPLES callout steer callers to pass `expectedSubject` when they know the subject, so shipping as-is.

packages/auth0-auth-js/examples/user-info.md
tusharpandey13: Fixed in 3f11703. The example now reads `error.statusCode` and `error.headers?.get('x-request-id')` instead of `error.cause?.error`, with a note that the underlying client throws before parsing the body so `cause.error` is always empty. Kept the `error.code` line.

packages/auth0-auth-js/src/auth-client.spec.ts
tusharpandey13: Removed in 3f11703. Confirmed the inner `afterEach` was redundant with the outer `server.resetHandlers()` and dropped it; full suite stays green.

packages/auth0-auth-js/src/auth-client.spec.ts
tusharpandey13: Addressed in 3f11703. Added A15, which returns a 401 with a `WWW-Authenticate` header and asserts `statusCode === 401` plus `headers.get('x-request-id')`, pinning the header-present path that Auth0 actually serves. The existing no-header cases continue to cover the `OperationProcessingError` path.

packages/auth0-auth-js/src/auth-client.spec.ts
tusharpandey13: Done in 3f11703. Folded A2/A4/A10 into A1 (all asserted `sub` against the same default handler). A10 is now a real caching test that counts `/.well-known/openid-configuration` fetches across two `getUserInfo` calls and asserts exactly one, using a unique domain so the shared discovery cache starts empty.

packages/auth0-auth-js/src/auth-client.spec.ts
tusharpandey13: Renamed in 3f11703 to "empty string expectedSubject is rejected, not silently skipped", with the comment corrected to describe the real mechanism: `?? skipSubjectCheck` forwards the empty string, and `oauth4webapi`'s `assertString` rejects it with `"expectedSubject" must not be empty` before any `sub` comparison. Still guards against a regression back to `||`.

packages/auth0-server-js/src/server-client.ts
tusharpandey13: Fixed in 3f11703. `ServerClient.getUserInfo` now has the documented `(options, storeOptions?, requestOptions?)` shape and forwards `requestOptions` to the auth-js method. Since it accepts `RequestOptions`, it stays out of the exclusion list; EXAMPLES.md now documents the third argument with a signal example. Thanks for confirming resolver mode.

packages/auth0-server-js/EXAMPLES.md
tusharpandey13: Removed the PR-introduced stray blank line in 3f11703. The TOC off-by-one and the 15-vs-14 `Passing StoreOptions` heading mismatch predate this PR, so leaving those for the separate doc-only PR as noted.

packages/auth0-server-js/EXAMPLES.md
tusharpandey13: Fixed in 3f11703, same treatment as the auth-js example: the server-js snippet now reads `error.statusCode` and `error.headers?.get('x-request-id')` instead of the always-empty `error.cause?.error`.

tusharpandey13:
Pushed 3f11703 addressing the review.

**Code**
- Public clients can now call `getUserInfo`: discovery no longer forces client auth for `/userinfo` (bearer-protected, needs none). A `clientAuthOptional` flag returns `client.None()` without poisoning the confidential-client auth memo or config cache.
- `getUserInfo` accepts a trailing `RequestOptions`; `ServerClient.getUserInfo` forwards it as the third arg.
- **Security (CodeRabbit):** `toOAuth2Error` now strips `set-cookie` from captured response headers so a session cookie can never leak onto a thrown `UserInfoError`. Inlined the delete rather than importing `filterSensitiveHeaders` to avoid an errors↔utils import cycle.
- `OperationProcessingError` (non-2xx without `WWW-Authenticate`, e.g. 429/5xx) carries no status/headers; documented on the JSDoc rather than fabricating values.

**Docs**
- Error examples read `error.statusCode` + `error.headers?.get('x-request-id')` instead of the always-empty `error.cause?.error`.
- Removed the PR-introduced stray blank line (MD028). Pre-existing TOC off-by-one deferred to a separate doc-only PR.

**Tests** (auth-js 542, server-js 323, all green)
- Folded redundant A2/A4/A10 into A1; A10 now pins discovery caching by fetch count.
- Renamed the empty-`expectedSubject` regression to reflect the real `assertString` rejection.
- Added public-client access, RequestOptions forwarding, WWW-Authenticate path, and set-cookie stripping coverage.

On `skipSubjectCheck`: keeping the default, since there is no subject to validate without a caller-supplied `expectedSubject`; dropped the node-auth0-parity framing in the reply.

packages/auth0-auth-js/src/auth-client.ts
```diff
@@ -526,39 +535,55 @@ export class AuthClient {
...
+      const configuration = await this.#createConfiguration(
+        entry.serverMetadata,
+        undefined,
+        clientAuthOptional
+      );
+      if (!clientAuthOptional) {
+        this.#serverMetadata = entry.serverMetadata;
+        this.#configuration = configuration;
+      }
+      return { configuration, serverMetadata: entry.serverMetadata };
```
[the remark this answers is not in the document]
tusharpandey13: Valid. Fixed in 798f090.

Root cause confirmed: `#inFlightDiscovery` was keyed by `cacheKey` alone. A strict discovery on a public client rejects at `#getClientAuth(false)` before any metadata is fetched, so a concurrent optional-auth `getUserInfo()` awaiting the same in-flight entry threw `MissingClientAuthError`.

Fix: the in-flight key now includes the auth mode (`strict`/`optional`), so the two callers never share a rejected promise. The successful, auth-independent server metadata is still shared across both via `#discoveryCache` once resolved, so we don't duplicate the metadata fetch.

Added regression test A16 overlapping `getServerMetadata()` (strict) with `getUserInfo()` (optional) on a public client: the strict call rejects with `MissingClientAuthError` while the optional call resolves.

packages/auth0-auth-js/src/auth-client.ts
```diff
@@ -526,39 +535,62 @@ export class AuthClient {
...
     }
 
-    const inFlight = this.#inFlightDiscovery.get(cacheKey);
+    // The in-flight map is keyed by auth mode as well as `cacheKey`. A strict
+    // discovery on a public client rejects at `#getClientAuth(false)` before any
+    // metadata is fetched; sharing that rejected promise with an optional-auth
+    // caller (`getUserInfo`) would wrongly fail a request this API supports. The
+    // successful, auth-independent server metadata is still shared via
+    // `#discoveryCache` once resolved.
+    const inFlightKey = `${cacheKey}|${clientAuthOptional ? 'optional' : 'strict'}`;
```
nandan-bhat: Keying the in-flight map by auth mode means two concurrent calls on a cold cache now fetch discovery twice, once for `strict` and once for `optional`. I fired `getUserInfo()` and `logout()` together on a fresh client and saw 2 requests to `/.well-known/openid-configuration` where it used to be 1.

This shows up more through `auth0-server-js`, since a handler that reads the profile while also refreshing or logging out is a normal request shape. It is worse in resolver mode, where a new `AuthClient` is built per call so the cache is cold often.

The reasoning in the comment is fair, but only the rejection needs to be isolated. The metadata itself does not depend on client auth. Could we key the in-flight fetch on `cacheKey` alone and let each caller build its own `Configuration` from the shared metadata?

packages/auth0-auth-js/src/auth-client.ts
```diff
@@ -526,39 +535,62 @@ export class AuthClient {
...
    */
-  async #discover() {
-    if (this.#configuration && this.#serverMetadata) {
+  async #discover(clientAuthOptional = false) {
+    // The strict instance cache (`this.#configuration`) is only ever built with
+    // real client authentication, so it may only satisfy strict callers. An
+    // `clientAuthOptional` caller (currently only `getUserInfo`, which needs no
+    // client auth) must never read or write it, otherwise a public-client
+    // `None()`-based configuration could poison a later confidential call.
+    if (!clientAuthOptional && this.#configuration && this.#serverMetadata) {
```
nandan-bhat: Agree a `None()` based configuration must never leak into a confidential call. The side effect is that the optional path gets no instance level caching at all, so it falls back to `#discoveryCache` on every call.

With `discoveryCache: { ttl: 0 }` on one client, 3 `getUserInfo()` calls made 3 discovery requests, while 3 strict calls made 1. So apps that turn the cache off pay a discovery round trip on every `getUserInfo()`.

Would a separate slot such as `#optionalConfiguration` work here? The metadata is already safe to share, only the auth bearing `Configuration` has to differ.

packages/auth0-auth-js/src/auth-client.ts
```diff
@@ -1670,6 +1701,77 @@ export class AuthClient {
...
+   * @param options Options including the access token and optional expected subject
+   *                for OIDC subject-consistency validation.
+   * @param requestOptions Optional per-request options (signal, headers, customFetch).
+   * @returns A Promise resolving to the user's profile claims.
+   * @throws {UserInfoError} If the /userinfo request fails (HTTP 401/403, network error,
+   *         subject mismatch if expectedSubject provided, or missing userinfo_endpoint).
+   *
+   * @remarks
+   * `/userinfo` is a bearer-protected resource and requires no client authentication, so this
+   * method works for public clients (no client secret, assertion, or mTLS): the supplied access
```
nandan-bhat, line 1719: Good that this works for public clients. I think it is worth saying here that it is currently the only method that does.

On a client built with just `domain` and `clientId` I could call `getUserInfo`, `mfa.listAuthenticators`, `mfa.challengeAuthenticator` and `passkey.challenge`, but `buildAuthorizationUrl`, `buildLogoutUrl` and `getServerMetadata` all throw `MissingClientAuthError`. Those three make no authenticated request, they only need discovery metadata, so they fail only because discovery is strict.

The token grants failing is correct, and this is not something this PR introduced. But `auth0-spa-js` builds a public client, so can we either reuse `clientAuthOptional` for those three in a follow up, or note the current limit here?

packages/auth0-server-js/src/server-client.ts
```diff
@@ -1038,6 +1038,39 @@ export class ServerClient<TStoreOptions = unknown> {
...
+    options: GetUserInfoOptions,
+    storeOptions?: TStoreOptions,
+    requestOptions?: RequestOptions
+  ): Promise<UserInfoResponse> {
+    const domain = await this.#resolveDomain(storeOptions);
+    const authClient = this.#getAuthClient(domain);
+    return authClient.getUserInfo(options, requestOptions);
```
nandan-bhat, line 1071: One note for resolver mode: `#getAuthClient` builds a fresh `AuthClient` on every call when `domain` is a function, so `getUserInfo` here never gets the instance cache and leans fully on the shared `discoveryCache`. That is fine while the cache is on, but it makes the two caching points I flagged in `auth-client.ts` more visible for multi tenant apps.

packages/auth0-auth-js/src/errors.ts
```diff
@@ -252,6 +271,16 @@ export class TokenRevocationError extends ApiError {
   }
 }
 
+/**
+ * Error thrown when retrieving user information from the /userinfo endpoint fails.
+ */
+export class UserInfoError extends ApiError {
```
nandan-bhat, line 277: `examples/http-requests-and-responses.md` says `statusCode` and `headers` are available on all error types and then lists them out. `UserInfoError` is missing from that list. Can we add it there so the doc stays accurate?

nandan-bhat:
@tusharpandey13 ,

This PR touches 2 packages:
- `auth0-auth-js`
- `auth0-server-js`

Under which package you are planning to `release` this PR ?


Also, the PR title & the first commit doesn't follow the commit convention.
It should be either 
- `feat(auth0-auth-js): add getUserInfo() support`
or
- `feat(auth0-server-js): add getUserInfo() support`

packages/auth0-auth-js/src/auth-client.ts
tusharpandey13: Fixed in 5b7b5a8. The in-flight discovery fetch is now keyed by `cacheKey` alone and the metadata fetch runs with `client.None()`, since discovery only reads the well-known document and never authenticates. Two concurrent callers on a cold cache now share a single `/.well-known/openid-configuration` request again. Only the auth-bearing `Configuration` is built per caller, so the strict-vs-optional distinction no longer touches the fetch. Added regression A17: a strict `getTokenByRefreshToken` racing an optional `getUserInfo` on a cold cache asserts exactly one discovery request.

packages/auth0-auth-js/src/auth-client.ts
tusharpandey13: Fixed in 5b7b5a8. Added a dedicated `#optionalConfiguration` instance slot. The `None()`-based configuration is still never shared with a confidential caller (strict uses `#configuration`, optional uses `#optionalConfiguration`), but the optional path now gets instance-level caching too. Added regression A18: with `discoveryCache: { ttl: 0 }`, three `getUserInfo()` calls now make a single discovery request (was three).

packages/auth0-auth-js/src/auth-client.ts
tusharpandey13: Documented in 5b7b5a8. Added a note on `getUserInfo` that it is currently the only method relaxing client auth for discovery, and that `buildAuthorizationUrl`/`buildLogoutUrl`/`getServerMetadata` remain strict on public clients today, with relaxing them tracked as a follow-up.

packages/auth0-server-js/src/server-client.ts
tusharpandey13: Acknowledged. Resolver mode builds a fresh `AuthClient` per call, so `getUserInfo` there leans on the shared `discoveryCache` rather than the instance cache. With the two `auth-client.ts` caching fixes in 5b7b5a8 (single shared in-flight fetch + `#optionalConfiguration`), the per-instance behavior is now correct; the shared cache remains the effective layer for resolver mode, which is expected. No code change here.

packages/auth0-auth-js/src/errors.ts
tusharpandey13: Fixed in 5b7b5a8. Added `UserInfoError` to the error-metadata list in `examples/http-requests-and-responses.md`, with the caveat that `statusCode`/`headers` are populated on the typical `WWW-Authenticate` 401/403 path but unset when the response omits the header (the `OperationProcessingError` path, e.g. 429/5xx).

tusharpandey13:
@nandan-bhat review `nits` and `lows` have been addressed
Post the fixes, the PR is now split into 2 PRs:
#266 : add `getUserInfo()` for auth-js
#267 : expose `getUserInfo()` for server-js

tusharpandey13:
Closing as we have package-spcific PRs ready, see above comment

PR #237, opened 2026-08-05
feat(auth0-server-js): send organization on the STT mint request

nandan-bhat:
## What this does
`requestSessionTransferToken` was not sending the `organization` parameter, so there was no way to mint an STT in an organization context. Auth0 accepts `organization` on the token exchange, we just were not passing it through.

This adds an optional `organization` to `RequestSessionTransferTokenOptions`, forwarded on the mint request.

## Key behavior
- `buildSessionTransferRedirect` already took an `organization`. That one scopes the target session, so both should be set when minting in an organization context. Passing it on the mint alone does not scope the session.
- A blank `organization` throws `OrganizationValidationError` before the agent session is read or refreshed. This matters because resolving the actor can refresh an expired session ID token, which is a network call plus a state store write. Failing after that on a client side validation error would rotate the caller's refresh token for a request that could never succeed.

## Scope
All changes are in `auth0-server-js`. `auth0-auth-js` already accepted `organization` on `exchangeToken`, so there is no lower layer change and no dependency bump. `auth0-fastify` passes options through and picks the option up on its next dependency bump.

## Not a breaking change

The option is optional, and calls that omit it send the same request as before. `organization` is on the core `PARAM_DENYLIST`, so anyone who tried passing it through `extra` was having it silently dropped and is unaffected.

## Testing

- Added tests for forwarding `organization` on the mint, omitting it when not provided, rejecting a blank value, and rejecting a blank value before the session refresh happens.
- Verified the tests are not vacuous by removing each guard and confirming a real failure.
- `auth0-server-js`: 408 tests pass. Build and lint clean across all 13 turbo tasks.
- Documented the mint vs redirect distinction in `EXAMPLES.md`, along with the actor token requirements (`RS256`/`PS256`, unexpired, issued to the same client) and that an explicit `actor` skips the session read entirely.

## Summary by CodeRabbit

- **New Features**
  - Session Transfer Token requests can now include an optional organization.
  - Organization context is forwarded during token creation and redirect flows, with validation for blank or invalid values.
  - Actor selection now supports automatic resolution from the agent session or explicit actor precedence.
  - Added validation requirements for actor tokens and documented behavior when no session lookup is needed.

- **Documentation**
  - Expanded examples and guidance for organization scoping, actor resolution, validation, and related errors.

### DIFF
--- a/packages/auth0-server-js/EXAMPLES.md
+++ b/packages/auth0-server-js/EXAMPLES.md
@@ -1230,7 +1230,7 @@ Because you are now supplying the token yourself, it is on you to supply a valid
 
 An ID token from the agent's own session on this client satisfies all of these. A token that fails any of them comes back as a `TokenExchangeError` from the token endpoint.
 
-If the customer belongs to an organization, pass it in **both** places — on the mint and on the redirect:
+If the customer belongs to an organization, `organization` can be passed on the mint, on the redirect, or both:
 
 ```ts
 const result = await serverClient.requestSessionTransferToken(
@@ -1247,7 +1247,9 @@ const url = serverClient.buildSessionTransferRedirect('https://app.example.com/a
 });
 ```
 
-The two serve different purposes. On the mint, the tenant validates the organization while issuing the STT, so a disallowed organization (or one the client's `organization_usage` forbids) fails at that call rather than later at the target's `/authorize`. On the redirect, it is what actually scopes the target session. Passing it on the mint alone does not scope the session, so set both. An `organization` the tenant rejects surfaces as a `TokenExchangeError`; a blank one throws `OrganizationValidationError` before the session is read or any network call is made.
+The two are separate parameters on separate requests. On the mint, the tenant validates the organization against the client's organization settings while issuing the STT, so an organization the client is not allowed to use fails at that call instead of the STT being issued without it. On the redirect, it is forwarded to the target's `/authorize` as part of a normal interactive login, the same as any other org-scoped login.
+
+The example above sets both, which is the safe default: the mint gets validated, and the target's login is explicitly org-scoped. An `organization` the tenant rejects surfaces as a `TokenExchangeError`; a blank one throws `OrganizationValidationError` before the session is read or any network call is made.
 
 > [!NOTE]
 > `buildSessionTransferRedirect` attaches a single-use credential to the URL, so `targetLoginUrl` **must be a trusted, app-controlled value** and use `https` (`http` is allowed only for the loopback hosts `localhost`, `127.0.0.1`, and `[::1]`). Never derive it from untrusted input such as a `returnTo` parameter, or the token could leak to an attacker-controlled host.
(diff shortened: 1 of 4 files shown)


packages/auth0-server-js/EXAMPLES.md
```diff
@@ -1218,16 +1221,35 @@ const result = await serverClient.requestSessionTransferToken(
...
+Because you are now supplying the token yourself, it is on you to supply a valid one. When the actor token type is the ID token URN — which it is by default — Auth0 validates the token, so it must be an **unexpired, asymmetrically-signed Auth0 ID token**:
+
+- Signed with **RS256 or PS256**. `HS256` is rejected because it uses a shared secret.
+- Unexpired, and carrying `sub`, `iss`, `exp`, and `iat`.
+- Issued to the **same client** that is making the exchange (its `aud` must be this client's ID).
+- Belonging to a user who still exists and is not blocked.
+
+An ID token from the agent's own session on this client satisfies all of these. A token that fails any of them comes back as a `TokenExchangeError` from the token endpoint.
+
+If the customer belongs to an organization, `organization` can be passed on the mint, on the redirect, or both:
```
kishore7snehil: This reword now lists passing `organization` on the mint alone as a valid option, and the same commit drops the "recorded on the issued STT" line from the type docs. If the org isn't carried on the STT, minting with `organization` but not forwarding it on the redirect validates the org at mint time but leaves the target session un-scoped, since the redirect `organization` is the one that reaches `/authorize`.

Was mint-only meant to read as sufficient on its own, or should we still steer people to set both when the session needs to be org-scoped?

packages/auth0-server-js/src/server-client.spec.ts
```diff
@@ -7980,6 +7980,137 @@ test('requestSessionTransferToken - forwards scope and extra params', async () =
...
+      expect.objectContaining({
+        organization: 'org_abc123',
+      })
+    );
+  } finally {
+    exchangeSpy.mockRestore();
+  }
+});
+
+test('requestSessionTransferToken - omits organization when not provided', async () => {
```
kishore7snehil, line 8017: This check passes whether `organization` is missing or set to `undefined`, so it doesn't really prove the parameter was left out. In practice the core only adds `organization` when it has a real value, so a blank one never reaches the request anyway. If we want the test to actually confirm that, we could check the request body itself, or use `expect.not.objectContaining({ organization: expect.anything() })`.

kishore7snehil (changes requested):

packages/auth0-server-js/EXAMPLES.md
nandan-bhat: Mint-only was not meant to look like enough on its own. I reworded it so the two are shown as doing different things, and I added a line saying the redirect one is what scopes the session, so pass it there too when you need that. I updated the type doc for organization the same way.

I have updated the docs in baa7fb87bf3c0378b8e4ad89c33c45bee59539cc

packages/auth0-server-js/src/server-client.spec.ts
nandan-bhat: `objectContaining({ organization: undefined })` does not match a missing key. I tested it. It only matches when the key is there and set to undefined. The real problem is that we always write organization: options.organization in the object, so the key is always present. The old check could only tell us `present and undefined`. It could never catch the key being dropped.

`expect.not.objectContaining({ organization: expect.anything() })`   is better. It passes when the key is missing or undefined, and fails on a real value, so it still works if we later stop sending the key.

Made changes in baa7fb87bf3c0378b8e4ad89c33c45bee59539cc

kishore7snehil (approved):
LGTM!

PR #239, opened 2026-08-09
fix(auth0-auth-js): send client credentials on passkey register/challenge

nandan-bhat:
## Why we are raising this PR
`register()` and `challenge()` were **broken for confidential clients**. Both methods send only `client_id` in the request body. Auth0 requires a **`client_secret`** in the body when the application is a confidential client, so every such call was rejected as an unauthorized client. Only public clients worked.

## What changed
`PasskeyClient` now sends the client credential that these two endpoints accept.

- New `buildClientAuthBody()` in `src/passkey/utils.ts` builds the client authentication fields for the request body. `register()` and `challenge()` both spread it in.
- `AuthClient` passes `clientSecret` and `useMtls` down to `PasskeyClient`.
- Behaviour by configuration:

| Configuration | Request body |
|---|---|
| Public client | `{ client_id }` |
| `clientSecret` | `{ client_id, client_secret }` |
| Empty `clientSecret` | `{ client_id }` |
| `useMtls` | `{ client_id }` |
| `useMtls` and `clientSecret` | `{ client_id }` |

`useMtls` takes **precedence** over `clientSecret`. Sending both would put a certificate and a body credential in the same request, which Auth0 rejects.

`getTokenByPasskey()` is **unchanged**. It goes through the token endpoint and already authenticated correctly.

## No breaking changes
Public clients send a **byte identical** request body to before this change. This was checked against the built `dist` through the public `AuthClient` API, so SPAs and regular web apps that work today keep working.

This PR also adds **zero public API**. `src/passkey/errors.ts` and `src/passkey/index.ts` are byte identical to `main`, and the new helper and its type do not appear in `dist/index.d.ts`.

## Why no new error class for private key JWT
These endpoints accept `client_secret` as their **only** body level credential. A client configured with only a `clientAssertionSigningKey` has no credential the SDK can send.

We deliberately **do not throw** an SDK side error for this case. If Auth0 later starts accepting `client_assertion` here, a client side refusal would block a configuration the server had begun supporting, and removing the error class would then be a **breaking change**. Adding public symbols to an SDK is easy, removing them is not. So the SDK sends what it can and lets Auth0's answer through unchanged. The limitation is **documented** in `EXAMPLES.md` and `README.md` instead.

The same reasoning covers clients that rely on mTLS alone. These two endpoints are not served on the mTLS endpoint aliases, so such a client also has no credential the SDK can send.

The SDK does **not** interpret, normalize, or document the error code Auth0 returns in either case, because it does not handle these errors. It puts the response on `error.cause` untouched. The docs describe that **shape** and tell readers to log it rather than branch on a specific code, since the code depends on how the application is configured in the Dashboard.

## Testing
- New tests cover, for **both** `register()` and `challenge()`: confidential client, public client, mTLS, empty secret, no `client_assertion` added, and the rejection when no acceptable credential is configured, surfaced as is.
- Integration tests in `auth-client.spec.ts` confirm `AuthClient` forwards the right options, and that an mTLS only client sends `client_id` alone to the regular host.

### DIFF
--- a/packages/auth0-auth-js/EXAMPLES.md
+++ b/packages/auth0-auth-js/EXAMPLES.md
@@ -139,7 +139,7 @@ const tokenResponse = await auth0.getTokenByClientCredentials({
 
 - **Client Certificate**: Your application must have a valid client certificate issued by a Certificate Authority (CA) that Auth0 trusts.
 - **Domain Configuration**: Your Auth0 tenant must be configured to support mTLS endpoints.
-- **No Additional Auth**: When `useMtls: true`, you don't need `clientSecret` or `clientAssertionSigningKey`.
+- **No Additional Auth**: When `useMtls: true`, you don't need `clientSecret` or `clientAssertionSigningKey`. The one exception is `passkey.register()` / `passkey.challenge()`, which are not served on the mTLS endpoint aliases and accept only a `clientSecret`. See [Using Passkeys](#using-passkeys).
 - **Custom Fetch Required**: You must provide a `customFetch` implementation that includes the client certificate in the TLS handshake.
 
 > [!IMPORTANT]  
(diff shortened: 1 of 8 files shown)


Piyush-85 (approved):
LGTM

PR #240, opened 2026-08-10
feat(auth0-auth-js): add anonymous sessions support

yogeshchoudhary147:
## Summary

- Adds `AnonymousSessionClient` sub-client exposed as `authClient.anonymous`
- Implements `POST /anonymous/token` (create + re-mint) and `POST /anonymous/logout`
- Includes full token renewal logic in `getTokenSilently()` — silently re-creates session on `session_expired`/`invalid_session_token`, surfaces all other errors
- Exports shared types (`AnonymousSession`, `AnonymousTokens`, `AnonymousSessionClaims`, `CreateAnonymousSessionOptions`, `GetAnonymousTokenSilentlyOptions`) and `AnonymousSessionError` for downstream SDKs (auth0-spa-js, auth0-server-js)

## Test plan

- [ ] 25 unit tests covering all methods and renewal paths pass (`npx vitest run src/anonymous-session/anonymous-session-client.spec.ts`)
- [ ] Full package test suite passes (`npx vitest run`)
- [ ] No TypeScript errors in new files

## Summary by CodeRabbit

* **New Features**
  * Added anonymous session support for creating, renewing, silently recovering, and logging out sessions.
  * Exposed anonymous session APIs, types, and structured error handling through the authentication client.
  * Added support for configuring audience, scope, metadata, and session credentials.
* **Tests**
  * Added comprehensive coverage for successful flows, expiration handling, fallback behavior, request details, and API errors.

### DIFF
--- a/packages/auth0-auth-js/src/anonymous-session/anonymous-session-client.spec.ts
+++ b/packages/auth0-auth-js/src/anonymous-session/anonymous-session-client.spec.ts
@@ -70,16 +70,7 @@ const restHandlers = [
     });
   }),
 
-  http.post(`https://${domain}/anonymous/logout`, async ({ request }) => {
-    const body = (await request.json()) as Record<string, unknown>;
-
-    if (body.session_token === 'unknown-session-token') {
-      return HttpResponse.json(
-        { error: 'invalid_session_token', error_description: 'Session token not found' },
-        { status: 400 }
-      );
-    }
-
+  http.post(`https://${domain}/anonymous/logout`, () => {
     return new HttpResponse(null, { status: 204 });
   }),
 ];
@@ -226,21 +217,57 @@ describe('createSession', () => {
       code: 'server_error',
     });
   });
+
+  test('throws AnonymousSessionError when expires_in is missing from response', async () => {
+    server.use(
+      http.post(`https://${domain}/anonymous/token`, () =>
+        HttpResponse.json({
+          access_token: accessToken,
+          token_type: 'Bearer',
+          // expires_in intentionally omitted
+          session_token: sessionToken,
+        })
+      )
+    );
+
+    const client = makeClient();
+    await expect(client.createSession()).rejects.toMatchObject({
+      code: 'server_error',
+    });
+  });
+
+  test('throws AnonymousSessionError with code metadata_too_large when metadata exceeds 1 KB', async () => {
+    const client = makeClient();
+    const largeMetadata = { data: 'x'.repeat(1025) };
+
+    await expect(client.createSession({ metadata: largeMetadata })).rejects.toMatchObject({
+      name: 'AnonymousSessionError',
+      code: 'metadata_too_large',
+    });
+  });
 });
 
-// ─── getToken ─────────────────────────────────────────────────────────────────
+// ─── getTokenSilently ─────────────────────────────────────────────────────────
 
-describe('getToken', () => {
-  test('re-mints access token using existing session token', async () => {
+describe('getTokenSilently', () => {
+  test('creates a new session when no sessionToken is provided', async () => {
     const client = makeClient();
-    const session = await client.getToken(sessionToken);
+    const session = await client.getTokenSilently();
+
+    expect(session.sessionToken).toBe(sessionToken);
+    expect(session.accessToken).toBe(accessToken);
+  });
+
+  test('re-mints access token when sessionToken is provided', async () => {
+    const client = makeClient();
+    const session = await client.getTokenSilently({ sessionToken });
 
     expect(session.accessToken).toBe('renewed-access-token');
-    expect(session.sessionToken).toBe(sessionToken); // session token unchanged
+    expect(session.sessionToken).toBe(sessionToken);
     expect(session.expiresAt).toBeGreaterThan(Math.floor(Date.now() / 1000));
   });
 
-  test('sends the session_token in the request body', async () => {
+  test('sends sessionToken, audience and scope in the request body', async () => {
     let capturedBody: Record<string, unknown> = {};
 
     server.use(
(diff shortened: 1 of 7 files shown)


packages/auth0-auth-js/src/anonymous-session/anonymous-session-client.ts
```diff
@@ -0,0 +1,325 @@
...
+    // No session — create one from scratch
+    if (!session) {
+      return this.createSession(options);
+    }
+
+    // Access token is still valid — nothing to do
+    const now = Math.floor(Date.now() / 1000);
+    if (session.expiresAt > now) {
+      return session;
+    }
```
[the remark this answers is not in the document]
yogeshchoudhary147: Intentionally skipped at this layer. `auth0-auth-js` is stateless and only makes HTTP calls — it has no opinion on UX or storage. Expiry leeway belongs in the higher-level SDKs (`auth0-spa-js`, `auth0-server-js`) that own session storage and the user-facing `getTokenSilently`. Adding it here would risk compounding with whatever leeway those layers apply.

packages/auth0-auth-js/src/anonymous-session/anonymous-session-client.ts
```diff
@@ -0,0 +1,325 @@
...
+      throw new AnonymousSessionError(
+        errorBody.error,
+        errorBody.error_description || 'Anonymous token request failed',
+        errorBody
+      );
+    }
+
+    const apiResponse = (await response.json()) as AnonymousTokenApiResponse;
+    return parseTokenResponse(apiResponse);
+  }
```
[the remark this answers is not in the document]
yogeshchoudhary147: The timeout concern is valid but out of scope for this PR. No other sub-client in this package (MfaClient, DatabaseClient, etc.) implements a timeout — they all rely on the caller's fetch implementation or platform defaults. If we add this, it should be a cross-cutting change applied uniformly across all sub-clients, not just here.

packages/auth0-auth-js/src/anonymous-session/errors.ts
```diff
@@ -0,0 +1,59 @@
...
+  constructor(code: AnonymousSessionErrorCode, message: string, cause?: AnonymousSessionApiErrorResponse) {
+    super(message);
+    this.name = 'AnonymousSessionError';
+    this.code = code;
+    this.cause = cause && {
+      error: cause.error,
+      error_description: cause.error_description,
+      message: cause.message,
+    };
+  }
```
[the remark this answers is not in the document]
yogeshchoudhary147: Intentionally consistent with the existing pattern in this package — MfaError, MyAccountError, and all other sub-client error classes declare `public cause` the same way. Changing only AnonymousSessionError would be inconsistent. If we align with ES2022 Error.cause, that should be a separate cross-cutting refactor across all error classes.

packages/auth0-auth-js/src/anonymous-session/types.ts
```diff
@@ -0,0 +1,146 @@
...
+export interface AnonymousSessionClientOptions {
+  /**
+   * The Auth0 domain (without https://).
+   * @example 'example.auth0.com'
+   */
+  domain: string;
+  /**
+   * The client ID of the application.
+   */
+  clientId: string;
```
nandan-bhat, line 13: This only takes `clientId`, so there is no way to pass a `clientSecret`. For confidential clients, which is the normal case for the server side SDKs like `auth0-server-js`, the `/anonymous/token` call will be rejected with `invalid_client` because no client authentication is sent. Per PRD Req 19, if a client is set up to need a secret, the endpoint expects it in the request. Can we accept `clientSecret` here and forward it from `AuthClient` the same way `mfa`, `passwordless` and `database` already do, then send it as HTTP Basic when it is present? Public clients can keep the current `client_id` in body behavior.

packages/auth0-auth-js/src/anonymous-session/anonymous-session-client.ts
```diff
@@ -0,0 +1,333 @@
...
+   * let session = loadSessionFromStorage();
+   *
+   * session = await authClient.anonymous.getTokenSilently(session, {
+   *   audience: 'https://api.example.com',
+   * });
+   *
+   * // session.accessToken is guaranteed to be valid at this point
+   * ```
+   */
+  async getTokenSilently(
```
nandan-bhat: This takes the whole `session` object and owns the create vs renew decision internally. That fits SPA nicely where this client can hold the session, but `auth0-server-js` keeps the session in its own store and does not hand the session object down to the core. On the server side it expects a stateless call like `getTokenSilently({ sessionToken, audience, scope })`. Since this core is shared by both the SPA and the server SDKs, can we settle on one shape before this merges? As it stands server-js cannot call this and has to fall back to `getToken`.

packages/auth0-auth-js/src/anonymous-session/anonymous-session-client.ts
```diff
@@ -0,0 +1,333 @@
...
+   * @param sessionToken - The session token of the anonymous session to end
+   * @returns Promise that resolves when the session has been ended
+   * @throws {AnonymousSessionError} When the request fails
+   *
+   * @example
+   * ```typescript
+   * await authClient.anonymous.logout(session.sessionToken);
+   * ```
+   */
+  async logout(sessionToken: string): Promise<void> {
```
nandan-bhat: `logout` requires a `sessionToken` string. In the cookie based RWA flow the `auth0_anon` cookie carries the session and `credentials: 'include'` does the work, so the caller may not actually have a token in hand. Could we make `sessionToken` optional to cover that case?

packages/auth0-auth-js/src/anonymous-session/anonymous-session-client.spec.ts
```diff
@@ -0,0 +1,470 @@
...
+    // Simulate feature_not_enabled
+    if (body.client_id === 'disabled-client') {
+      return HttpResponse.json(
+        { error: 'feature_not_enabled', error_description: 'Anonymous sessions are not enabled for this tenant' },
+        { status: 403 }
+      );
+    }
+
+    // Simulate metadata_too_large
+    if (body.metadata && JSON.stringify(body.metadata).length > 1024) {
```
nandan-bhat, line 46: The mock handles `metadata_too_large` but I could not find a test that actually sends more than 1KB of metadata to reach this branch. Since this is one of the errors the spec calls out by name, can we add a test that sends a large metadata payload and asserts the `code` is `metadata_too_large`?

packages/auth0-auth-js/src/anonymous-session/types.ts
```diff
@@ -0,0 +1,146 @@
...
+  sessionToken?: string;
+}
+
+/**
+ * Decoded claims from an anonymous session token (JWT).
+ *
+ * Note: In EA, session tokens are issued as JWE (opaque). This type is reserved
+ * as an extension point for JWT format support that may be added in the future.
+ */
+export interface AnonymousSessionClaims {
```
nandan-bhat, line 90: This type is exported but nothing in the client produces or returns it today. There is no `getClaims` and no transparent JWT token handling yet. Keeping it as a placeholder is fine, but can we make the doc clearer that no method returns it right now, so consumers do not assume claim reading is supported?

packages/auth0-auth-js/src/anonymous-session/anonymous-session-client.ts
```diff
@@ -0,0 +1,333 @@
...
+ */
+const SESSION_INVALIDATION_CODES = new Set(['session_expired', 'invalid_session_token']);
+
+/**
+ * Parses an Auth0 token API response into an {@link AnonymousTokens} object.
+ */
+function parseTokenResponse(apiResponse: AnonymousTokenApiResponse): AnonymousTokens {
+  const now = Math.floor(Date.now() / 1000);
+  const expiresIn = apiResponse.expires_in;
+  if (typeof expiresIn !== 'number' || !Number.isFinite(expiresIn)) {
```
nandan-bhat, line 27: Nice, good to validate `expires_in` is a real number before computing `expiresAt`. This keeps a `NaN` expiry from leaking into the session.

packages/auth0-auth-js/src/anonymous-session/types.ts
yogeshchoudhary147: Fixed! Added `clientSecret` to `AnonymousSessionClientOptions` and it gets forwarded from `AuthClient` the same way `mfa`, `passkey` and `database` already do. Sent as `client_secret` in the request body when present.

packages/auth0-auth-js/src/anonymous-session/anonymous-session-client.ts
yogeshchoudhary147: Good point. Reshaped `getTokenSilently` to a stateless signature `(options?: GetAnonymousTokenSilentlyOptions)` where `sessionToken` is just an option alongside `audience` and `scope`. Server-side SDKs can now pass whatever token they have in their store. Also moved the expiry check out of core since higher layers like `auth0-server-js` own that state.

packages/auth0-auth-js/src/anonymous-session/anonymous-session-client.ts
yogeshchoudhary147: Went a bit further on this one. Checked the product spec and it consistently shows `logout()` with no args across all examples. Session is identified purely via the `auth0_anon` cookie through `credentials: 'include'`, so the caller never needs a token in hand. Removed `sessionToken` from the signature entirely and the request body now only has `client_id`.

packages/auth0-auth-js/src/anonymous-session/anonymous-session-client.spec.ts
yogeshchoudhary147: Added a test that sends metadata with a 1025-char value and asserts the code is `metadata_too_large`.

packages/auth0-auth-js/src/anonymous-session/types.ts
yogeshchoudhary147: Updated the JSDoc to make it clear no method returns this type in EA. Kept the export for forward compatibility so there is no breaking change when session tokens move to a readable JWT format.

packages/auth0-auth-js/src/anonymous-session/anonymous-session-client.ts
yogeshchoudhary147: Thanks\! Also added a test for the case where `expires_in` is missing from the response.

nandan-bhat (approved):

PR #242, opened 2026-08-10
fix(auth0-server-js): require auth0-auth-js ^1.12.1 for passkey client auth

nandan-bhat:
## Description
`/passkey/register` and `/passkey/challenge` accept `client_secret` as their only body-level client credential. `@auth0/auth0-auth-js` only started sending it in **1.12.1** ([#239](https://github.com/auth0/auth0-auth-js/pull/239)), so on 1.12.0 confidential clients could not authenticate against either endpoint.

No source changes are needed in this package. `ServerClient` already forwards `clientSecret` and `useMtls` into its `AuthClient` options, so the fix propagates through both the static and resolver (multi-tenant) paths untouched. This PR raises the dependency floor and documents the constraint.

- **`package.json`**: `@auth0/auth0-auth-js` `^1.12.0` → `^1.12.1`. A fresh install already resolves 1.12.1 under the old caret, but a consumer with a lockfile pinned to 1.12.0 keeps getting the broken version. Only the floor moves them.
- **JSDoc** on `register()` / `challenge()`: these endpoints do not accept a private key JWT and are not served on the mTLS endpoint aliases, so a client configured with only `clientAssertionSigningKey` or only `useMtls` has no credential the SDK can send. `getToken()` is unaffected, since it goes through the token endpoint.
- **`EXAMPLES.md`**: same explanation, cross-linked from both challenge sections.
- **9 regression tests** asserting the request body on the wire for public, client-secret, mTLS, and private-key-JWT-only clients, plus resolver mode. They fail against `auth0-auth-js@1.12.0` and pass against 1.12.1, so they cover the defect rather than the plumbing.

## Testing
Verified against the **published** `@auth0/auth0-auth-js@1.12.1` tarball, not just workspace source: all 9 passkey assertions pass. Confirmed `buildClientAuthBody` is in the shipped `dist` and is spread into both request bodies.

### DIFF
--- a/packages/auth0-server-js/EXAMPLES.md
+++ b/packages/auth0-server-js/EXAMPLES.md
@@ -355,7 +355,7 @@ const authorizationUrl = await serverClient.startInteractiveLogin();
 
 - **Client Certificate**: Your application must have a valid client certificate issued by a Certificate Authority (CA) that Auth0 trusts.
 - **Domain Configuration**: Your Auth0 tenant must be configured to support mTLS endpoints.
-- **No Additional Auth**: When `useMtls: true`, you don't need `clientSecret` or `clientAssertionSigningKey`.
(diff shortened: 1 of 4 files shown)


Piyush-85 (approved):

Piyush-85:
LGTM

PR #236, opened 2026-08-04
feat(auth0-server-js): forward HTTP response metadata on success and error paths

tusharpandey13:
## Summary

Mirrors the HTTP response/error metadata surface added to `@auth0/auth0-auth-js` in #234 so that `@auth0/auth0-server-js` consumers can access the same metadata (`status`, `statusText`, `headers` on success; `statusCode`, `body`, `headers` on error). This restores parity with node-auth0 for observability (trace IDs, rate-limit headers) after the auth-separation split.

**Stacked on #234** (base branch `feat/response-error-metadata-authjs`). Merge #234 first.

## Why

node-auth0 exposed HTTP status, response body, and response headers on both success and error paths. The split SDKs lost this. #234 restored it for `auth0-auth-js`; this PR forwards it through `auth0-server-js`.

## What changed

**auth0-auth-js (base fix, 1 line):**
- The `@internal` tag removal from `TokenResponse.httpResponse` now lives in parent PR #234 (commit 7103a32). This PR no longer carries it as a duplicate.

**auth0-server-js (success path):**
- `TokenSet` and `ConnectionTokenSet` gain an optional `httpResponse?: HttpResponseMetadata` (re-exported from auth-js).
- `getAccessToken()` and `getAccessTokenForConnection()` forward the metadata captured by auth-js on a **fresh** token request.
- `requestSessionTransferToken()` now exposes `httpResponse` on `SessionTransferTokenResult`, consistent with other token-returning methods.

**auth0-server-js (error path):**
- Token delegation is not wrapped in try/catch, so the auth-js `ApiError` propagates unwrapped with `statusCode`/`body`/`headers` intact. Verified, no code change required.
- `#resolveSessionTransferActor` intentionally suppresses the original refresh error metadata in favor of a uniform `actor_unavailable` error; this is now documented with an inline comment.

**Sub-clients (MFA / Passkey / Database):**
- These return the auth-js response objects verbatim, so `httpResponse` forwards automatically. Tests now assert this contract explicitly rather than relying on incidental pass-through.

**Documentation:**
- EXAMPLES.md HTTP response metadata section now documents that session-establishment methods (`completeInteractiveLogin`, `loginBackchannel`, `completePasswordless`, `loginWithCustomTokenExchange`) intentionally do not expose `httpResponse`. Callers needing response metadata should use `getAccessToken`, `getAccessTokenForConnection`, `customTokenExchange`, or `requestSessionTransferToken`.

## Design decisions

[...]

PR #234, opened 2026-08-03
feat(auth0-auth-js): expose HTTP response metadata on success and error paths

tusharpandey13:
Exposes HTTP response metadata on the success and error paths for better observability.

**Changes:**
- TokenResponse: added `httpResponse?: { status, statusText, headers }`
- OAuth2Error + descendants: added `statusCode?`, `headers?`, `body?`
- All success paths attach metadata from captured fetch
- All error paths enrich with captured HTTP metadata + body
- 80+ tests asserting metadata presence and correctness

**Review findings addressed:**
- [H] Remove @internal tag from TokenResponse.httpResponse — was causing stripInternal to remove field from published .d.ts, breaking TS consumers (TS2339). Moved here from #236 for self-consistency.
- [D] Extract attachHttpResponseMetadata helper — replaced 12 copy-pasted 6-line blocks across success paths
- [G] Deduplicate error metadata assignment — toOAuth2Error extracts from openid-client errors (needed by passkey/passwordless clients), enrichOAuth2ErrorWithMetadataAndBody now only sets if undefined to avoid redundant overwrites
- [E] Remove inconsistent optional chaining on getCapturedResponse (L421,456) — #buildRequestFetch always returns CapturingFetch, so `?.` was misleading
- [A] Remove dead composeRequestFetch_impl no-op function
- [C] Inline enrichOAuth2ErrorWithMetadata into its sole caller
- Test fixes: vitest `fail()` → `expect.fail()`, PromiseSettledResult .reason access guarded, mock.calls[0] destructuring replaced with .at(0) for TS strictness

**Correctness:**
- ✅ 389 tests passing (parent #230 added 2 mTLS tests, now included via rebase)
- ✅ tsc, build, lint green
- ✅ Rebased onto parent #230 new tip 6079d5f (mTLS fix)
- ✅ Test coverage: success/error metadata, concurrent calls, isolation, optional chaining, request-scoped options

PR #244, opened 2026-08-11
feat(auth0-server-js): add per-request options (signal, headers, customFetch)

tusharpandey13:
Mirrors #230 into `auth0-server-js`: `RequestOptions` (`signal`, `headers`, `customFetch`) as a trailing positional parameter on every method that performs a network call.

Based on #230 rather than `main`, so the diff stays limited to the server-js side. GitHub will retarget it to `main` once #230 merges. This is a separate PR, not part of the #230 → #234 → #236 chain.

## What changed

`requestOptions` is threaded through 15 `ServerClient` methods and all 9 sub-client methods, always as the last argument, always optional. It applies to a single call and never mutates client configuration.

Four behaviors are worth calling out, and each is documented in `EXAMPLES.md` and in TSDoc at the method:

- **Cache hits are a no-op.** `getAccessToken` and `getAccessTokenForConnection` return from the state store before any request. A caller passing a `signal` to bound the total call time will find it has no effect on that path.
- **`logout` forwards to the revocation only.** Building the logout URL is local string work.
- **The MFA argument shape differs.** `listAuthenticators`, `enrollAuthenticator` and `challengeAuthenticator` never touch the store, so they take `(options, requestOptions?)`. Everything else takes `(options, storeOptions?, requestOptions?)`. This matches auth-js.
- **`getAccessToken` extends the options overload only.** The legacy store-options-only overload is slated for removal in the next major, so `getAccessToken(undefined, { signal })` does not compile. Confirmed with a real `tsc` run, not by inspection.

`getUser`, `getSession`, `handleBackchannelLogout`, the `start*` URL builders and `buildSessionTransferRedirect` are deliberately excluded. `getUser` gets an explicit rationale in the docs because the upstream migration notes list it beside three methods that do accept the parameter, so readers will come looking: it is a pure state-store read, so the parameter could never do anything, and a dead parameter costs more than the asymmetry. `handleBackchannelLogout` is excluded because auth-js `verifyLogoutToken` takes no `requestOptions`.

## Removed test scaffolding from published source

The first commit deletes test-only code that had been left in shipping files. The database and passkey sub-clients each carried a public `authClient` getter returning a `Proxy` that called `getAuthClient('auth0.local')`, hardcoding a fake domain into the published package, alongside a `#cacheAuthClient` method and a `#cachedDomain` field that was written and never read. `ServerMfaClient` had a redundant getter of its own, marked `@internal` but exported regardless, duplicating what `ServerClient.authClient` already provides.

Nothing outside those files referenced any of it, and the caching constructed a second `AuthClient` on every sub-client call. The dead field also tripped `no-unused-private-class-members`, so lint was red before this change.

## Tests

472 passing, up from 415 with 29 failing. The affected tests were reworked to assert on the request MSW actually received, so they prove the options reach the wire rather than that an internal call happened.

Two things found while fixing them:

The passkey per-request block was four comment lines asserting the behavior was "integration-tested via existing end-to-end tests". It was not tested at all. It is now 14 tests across `register`, `challenge` and `getToken`, covering headers, `customFetch`, an aborted signal, the omitted-options path and resolver mode. The file's MSW handlers had also never matched anything: they were registered on `/webauthn/*` rather than the `/passkey/*` paths the client calls, returned camelCase bodies where the API returns snake_case, and used a grant type the SDK does not send. `onUnhandledRequest` is now `error` so a future path drift fails loudly.

Two MFA tests passed a `credentialType` option to `listAuthenticators` that `ListAuthenticatorsOptions` does not declare, so they were asserting against an API that does not exist. This was invisible because nothing typechecks the specs: `build` is `tsup` over entry points, `test` is `vitest`, and CI runs neither `tsc` nor anything equivalent, even though `tsconfig.json` includes `src/**/*`. All 11 pre-existing type errors in this package are fixed here; adding a `typecheck` script and CI step is left for a follow-up, since it will surface errors in the other packages too.

The suite was verified by sabotage rather than by reading it: stripping every forwarded `requestOptions` from `ServerClient` turns 31 tests red, and restoring it returns all 472 to green.

## Gates

```
npm run build     7/7 pass
npm test          server-js 472/472, auth-js 387, api-js 274
npm run lint      0 errors
npx tsc --noEmit  0 errors
```

Additive and backward compatible: every new parameter is optional, and the no-options path is covered by tests.

---

Rebased onto updated #230 (includes mTLS endpoint-alias preservation fix). No server-js code changes required; the fix propagates via the consumed AuthClient. This PR will retarget to main once #230 merges.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


tusharpandey13:
### Update — parent/`main` integration

**Head:** `41b91d9`

No review findings were open on this PR. It inherits the #230 `AuthClient` fixes (mTLS endpoint-alias preservation) transitively through the consumed dependency.

**Integration:**
- Merged the updated base #230 (which carries `main`'s anonymous sessions #240, release v1.12.1, and the server-js passkey client-auth change #242). The merge was a clean auto-union — no manual conflict resolution required. The server-js passkey client-auth changes (#242) and this PR's per-request-options threading (`signal`/`headers`/`customFetch`) both coexist on the passkey register/challenge/getToken methods.

**Gate:** auth0-server-js 481 tests, auth0-auth-js 413 tests, build + lint green, tsc clean.

This PR retargets to `main` once #230 merges.

PR #247, opened 2026-08-20
fix(auth0-auth-js): remove client_secret from logout request body

yogeshchoudhary147:
## Summary

`POST /anonymous/logout` only accepts `client_id` in the request body. The `logout()` method was incorrectly including `client_secret`, causing a 400 for confidential clients.

Removed the `client_secret` block from `logout()`. The `#postAnonymousToken` method correctly continues to send `client_secret` to `POST /anonymous/token` where it is required for confidential clients.

## Test plan

- All existing tests pass (404 tests)

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved anonymous-session logout requests by preventing unnecessary client-secret data from being included.
  * Clarified that logout must run in a browser to clear the anonymous-session cookie; server-side requests may return successfully without revoking the session.

### DIFF
--- a/packages/auth0-auth-js/src/anonymous-session/anonymous-session-client.ts
+++ b/packages/auth0-auth-js/src/anonymous-session/anonymous-session-client.ts
@@ -245,6 +245,13 @@ export class AnonymousSessionClient {
    * The anonymous session is identified via the `auth0_anon` cookie, which is
    * sent automatically through `credentials: 'include'`.
    *
+   * **This method must be called from a browser.** The `auth0_anon` cookie is
+   * set on the Auth0 domain and is HttpOnly — it can only be cleared by Auth0
+   * in response to this request. If not cleared, the cookie survives and is
+   * sent to `/authorize` on the next login, re-attaching the anonymous session
+   * to the authenticated user. Calling from a server-side environment (where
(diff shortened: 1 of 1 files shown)


packages/auth0-auth-js/src/anonymous-session/anonymous-session-client.ts
```diff
@@ -263,10 +263,6 @@ export class AnonymousSessionClient {
       client_id: this.#clientId,
     };
 
-    if (this.#clientSecret) {
```
nandan-bhat: The removal is correct and complete, `#clientSecret` is still used by `#postAnonymousToken` so nothing goes dead.

One thing though. The docstring for this method a few lines above still says it invalidates the session on Auth0's side, and that the session is identified by the `auth0_anon` cookie sent through `credentials: 'include'`. Neither is true on a server runtime, since there is no cookie jar there and the endpoint revokes nothing.

Before this change a confidential client got a `400` and at least knew something was wrong. After it, the call returns `204` and does nothing, while the docs still promise the session was ended. That is a harder thing to debug.

Can we update the text to say the endpoint only clears the browser cookie, revokes no tokens, and is effectively a no-op outside a browser?

yogeshchoudhary147:

PR #251, opened 2026-08-20
fix(auth0-auth-js): support private_key_jwt client authentication

yogeshchoudhary147:
## What

Anonymous session token requests only supported `client_secret_post`. Confidential clients configured for `private_key_jwt` had no way to authenticate — they were locked out entirely.

## Changes

- `AnonymousSessionClientOptions` gains `clientAssertionSigningKey` and `clientAssertionSigningAlg`
- New `anonymous-session/utils.ts` with `buildClientAuthBody` — resolution order: `private_key_jwt` → `client_secret_post` → public client (no body auth)
- `AnonymousSessionClient#postAnonymousToken` calls it instead of the bespoke secret-only block
- `AuthClient` passes the new fields through when constructing the anonymous client

## Notes

- Public clients (no auth configured) return an empty auth body — valid for this endpoint, unlike the passwordless variant which throws `MissingClientAuthError`
- mTLS is transport-level and remains handled via `customFetch`; not included here per the gaps analysis

## Summary by CodeRabbit

* **New Features**
  * Added support for multiple authentication methods when starting anonymous sessions, including client secrets and private key JWTs.
  * Added configuration for signing keys and algorithms, with RS256 used by default.
  * Public clients can now authenticate without client credentials.
* **Bug Fixes**
  * Logout requests no longer send client secrets.

### DIFF
--- a/packages/auth0-auth-js/src/anonymous-session/anonymous-session-client.spec.ts
+++ b/packages/auth0-auth-js/src/anonymous-session/anonymous-session-client.spec.ts
@@ -1,8 +1,23 @@
 import { expect, test, describe, beforeAll, afterAll, afterEach, vi } from 'vitest';
 import { setupServer } from 'msw/node';
 import { http, HttpResponse } from 'msw';
+import { decodeJwt, decodeProtectedHeader } from 'jose';
 import { AnonymousSessionClient } from './anonymous-session-client.js';
 import { AnonymousSessionError } from './errors.js';
+
+const exportPrivateKeyToPem = async (privateKey: CryptoKey): Promise<string> => {
+  const pkcs8 = await crypto.subtle.exportKey('pkcs8', privateKey);
+  const keyBase64 = Buffer.from(pkcs8).toString('base64');
+  const keyLines = keyBase64.match(/.{1,64}/g) ?? [keyBase64];
+  return `-----BEGIN PRIVATE KEY-----\n${keyLines.join('\n')}\n-----END PRIVATE KEY-----`;
+};
+
+const generateRsaKeyPair = () =>
+  crypto.subtle.generateKey(
+    { name: 'RSASSA-PKCS1-v1_5', modulusLength: 2048, publicExponent: new Uint8Array([1, 0, 1]), hash: { name: 'SHA-256' } },
+    true,
+    ['sign', 'verify']
+  ) as Promise<CryptoKeyPair>;
 
 const domain = 'auth0.local';
 const clientId = 'test-client-id';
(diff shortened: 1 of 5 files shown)


packages/auth0-auth-js/src/anonymous-session/utils.ts
```diff
@@ -0,0 +1,66 @@
...
+ * Resolution order: `private_key_jwt` → `client_secret_post` → public client
+ * (no body auth).
+ *
+ * Unlike the passwordless variant, an empty object is returned for a public
+ * client (no auth options configured): the anonymous token endpoint does not
+ * require client authentication for public clients.
+ *
+ * @internal
+ */
+export async function buildClientAuthBody(
```
nandan-bhat: This is the third copy of `buildClientAuthBody` in the package, after `passkey/utils.ts:26` and `passwordless/utils.ts:43`. Same name, same three constants, same claim building block, and the same `Claims mirror node-auth0 client-authentication` comment.

Copy three has already drifted from the other two on its first commit, since it has no `useMtls` handling and no `MissingClientAuthError`. That drift is the bug I flagged below.

I am not asking you to refactor all three in this PR. Can we raise a follow up to pull them into one shared helper though? Otherwise the next endpoint makes it four.

packages/auth0-auth-js/src/anonymous-session/utils.ts
```diff
@@ -0,0 +1,66 @@
...
+        ? options.clientAssertionSigningKey
+        : await importPKCS8(options.clientAssertionSigningKey as string, alg);
+
+    // Claims mirror node-auth0 client-authentication: iss/sub = clientId,
+    // aud = `https://{domain}/` (trailing slash), short-lived, unique jti.
+    const clientAssertion = await new SignJWT({})
+      .setProtectedHeader({ alg })
+      .setIssuer(clientId)
+      .setSubject(clientId)
+      .setAudience(`https://${domain}/`)
```
nandan-bhat: Noting this so nobody tidies it up later. The trailing slash is required and the value has to be the issuer identifier. `/anonymous/token` is stricter here than `/oauth/token`, it accepts only the issuer, and there is a server test for it at `anonymous_token_client_authentication.test.js:297` which rejects an assertion whose `aud` is the token endpoint URL.

Might be worth saying that in the comment above, since the rule is different from the other endpoints we talk to.

packages/auth0-auth-js/src/anonymous-session/anonymous-session-client.spec.ts
```diff
@@ -436,3 +451,110 @@ describe('expiresAt', () => {
...
+    expect(capturedBody).not.toHaveProperty('client_assertion');
+  });
+
+  test('sends no auth fields for a public client', async () => {
+    let capturedBody: Record<string, unknown> = {};
+
+    server.use(
+      http.post(`https://${domain}/anonymous/token`, async ({ request }) => {
+        capturedBody = (await request.json()) as Record<string, unknown>;
+        return HttpResponse.json({
```
nandan-bhat: Heads up on a clash with #252. That PR adds `session_expires_in` to the six existing handlers in this file and starts reading it in `parseTokenResponse`. The four handlers this PR adds do not return it.

Once both land, these four tests run with `session_expires_in` absent, so `sessionExpiresAt` comes out as `NaN` and the tests still pass, because nothing asserts on it. Whichever of the two merges second should add the field to these four.

packages/auth0-auth-js/src/anonymous-session/utils.ts
```diff
@@ -0,0 +1,66 @@
...
+ * require client authentication for public clients.
+ *
+ * @internal
+ */
+export async function buildClientAuthBody(
+  options: ClientAuthOptions,
+  clientId: string,
+  domain: string
+): Promise<Record<string, string>> {
+  if (options.clientAssertionSigningKey) {
```
nandan-bhat, line 35: This drops the `useMtls` branch that both of the other copies have. `passwordless/utils.ts:49` and `passkey/utils.ts:27` both return `{}` when `useMtls` is set, so no body level auth is sent at all.

Without it we send `client_secret` on a connection that is already authenticating with a certificate, and that counts as two authentication methods. I checked it against the API and it comes back as `400` with `error: invalid_request` and `error_description: 'Multiple authentication methods provided, please use one.'`

`AuthClient` passes `clientSecret` to this client unconditionally, so a client configured with `useMtls: true` that also has a secret on file will fail on every `/anonymous/token` call, while passkey and passwordless keep working fine.

Can we add the same early return here, and add `useMtls` to the `Pick` on line 14?

packages/auth0-auth-js/src/anonymous-session/utils.ts
nandan-bhat: This is the third copy of `buildClientAuthBody` in the package, after `passkey/utils.ts:26` and `passwordless/utils.ts:43`. Same name, same three constants, same claim building block, and the same `Claims mirror node-auth0 client-authentication` comment.

Copy three has already drifted from the other two on its first commit, since it has no `useMtls` handling and no `MissingClientAuthError`. That drift is the bug I flagged below.

I am not asking you to refactor all three in this PR. Can we raise a follow up to pull them into one shared helper though? Otherwise the next endpoint makes it four.

packages/auth0-auth-js/src/anonymous-session/utils.ts
nandan-bhat, line 48: Noting this so nobody tidies it up later. The trailing slash is required and the value has to be the issuer identifier. This endpoint is stricter here than `/oauth/token`, it takes the issuer only, and an assertion audienced to the token endpoint URL gets rejected. I confirmed that against the API.

Might be worth saying so in the comment above, since the rule is different from the other endpoints we talk to.

packages/auth0-auth-js/src/auth-client.ts
```diff
@@ -390,6 +390,8 @@ export class AuthClient {
       domain: this.#options.domain,
       clientId: this.#options.clientId,
       clientSecret: this.#options.clientSecret,
+      clientAssertionSigningKey: this.#options.clientAssertionSigningKey,
+      clientAssertionSigningAlg: this.#options.clientAssertionSigningAlg,
```
nandan-bhat, line 394: `useMtls` is missing here. The passkey block on line 341 and the passwordless block on line 373 both pass it. Without it the anonymous client cannot tell an mTLS client apart from a `client_secret_post` one, which is the `400` I described in `utils.ts`.

Both of the sibling blocks also carry a short comment explaining which auth options that sub client gets and why. Worth doing the same here, because the anonymous endpoint is the one place where a public client with no credentials at all is valid, and that is not obvious from the code.

packages/auth0-auth-js/src/anonymous-session/anonymous-session-client.spec.ts
nandan-bhat, line 486: Heads up on a clash with #252. That PR adds `session_expires_in` to the six existing handlers in this file and starts reading it in `parseTokenResponse`. The four handlers this PR adds do not return it.

Once both land, these four tests run with `session_expires_in` absent, so `sessionExpiresAt` comes out as `NaN` and the tests still pass, because nothing asserts on it. Whichever of the two merges second should add the field to these four.

yogeshchoudhary147:

PR #252, opened 2026-08-20
fix(auth0-auth-js): surface session_expires_in as sessionExpiresAt

yogeshchoudhary147:
## What

The platform returns `session_expires_in` on every `/anonymous/token` response (create and re-mint). The SDK was dropping it — `AnonymousTokenApiResponse` did not declare it, so `parseTokenResponse` never read it. Nothing on `AnonymousSession` described when the session token itself expired.

## Why it matters

- The session token is minted once and never reissued. `session_expires_in` counts down on each response, it does not reset.
- `auth0-server-js` picks a cookie `Max-Age` from this value. Without it, it falls back to 30 days from creation time — which is wrong on any tenant that shortens the anonymous session lifetime.
- `auth0-spa-js` has no way to know or act on session expiry.

## Changes

- `AnonymousTokenApiResponse.session_expires_in` — added as required `number` (platform always sends it)
- `AnonymousTokens.sessionExpiresAt` — added as required `number` (always computed)
- `AnonymousSession.sessionExpiresAt` — added as optional `number` (optional on public surface for forward-compat)
- `parseTokenResponse` — computes `sessionExpiresAt = now + session_expires_in`
- `createSession` and `#mintToken` — pass `sessionExpiresAt` through to the returned session
- Tests — 3 new cases: set on create, set on re-mint, re-mint value is less than create value (counts down)

### DIFF
--- a/packages/auth0-auth-js/src/anonymous-session/anonymous-session-client.ts
+++ b/packages/auth0-auth-js/src/anonymous-session/anonymous-session-client.ts
@@ -33,6 +33,7 @@ function parseTokenResponse(apiResponse: AnonymousTokenApiResponse): AnonymousTo
     expiresAt: now + expiresIn,
     scope: apiResponse.scope,
     sessionToken: apiResponse.session_token,
+    sessionExpiresAt: now + apiResponse.session_expires_in,
   };
 }
 
@@ -153,6 +154,7 @@ export class AnonymousSessionClient {
       sessionToken: tokens.sessionToken,
       accessToken: tokens.accessToken,
       expiresAt: tokens.expiresAt,
+      sessionExpiresAt: tokens.sessionExpiresAt,
       scope: tokens.scope,
     };
   }
@@ -234,6 +236,7 @@ export class AnonymousSessionClient {
       sessionToken,
       accessToken: tokens.accessToken,
       expiresAt: tokens.expiresAt,
+      sessionExpiresAt: tokens.sessionExpiresAt,
       scope: tokens.scope,
     };
   }
--- a/packages/auth0-auth-js/src/anonymous-session/types.ts
+++ b/packages/auth0-auth-js/src/anonymous-session/types.ts
@@ -42,6 +42,16 @@ export interface AnonymousSession {
    */
   expiresAt: number;
   /**
+   * Unix timestamp (seconds) at which the session token itself expires.
+   *
+   * The session token is minted once and never reissued, so this value counts
+   * down on every response rather than resetting. Present on both create and
+   * re-mint responses when the tenant has anonymous sessions enabled.
+   *
+   * Use this — not `expiresAt` — to drive cookie `Max-Age` or session expiry UI.
+   */
+  sessionExpiresAt: number;
+  /**
    * Scopes granted to this anonymous session.
    */
   scope?: string;
@@ -76,6 +86,11 @@ export interface AnonymousTokens {
    * Only present when a new session is created (not on re-mint).
    */
   sessionToken?: string;
+  /**
+   * Unix timestamp (seconds) at which the session token itself expires.
+   * Computed from `session_expires_in` in the API response.
+   */
+  sessionExpiresAt: number;
 }
 
 /**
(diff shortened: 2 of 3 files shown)


packages/auth0-auth-js/src/anonymous-session/types.ts
```diff
@@ -41,6 +41,16 @@ export interface AnonymousSession {
...
+  /**
+   * Unix timestamp (seconds) at which the session token itself expires.
+   *
+   * The session token is minted once and never reissued, so this value counts
+   * down on every response rather than resetting. Present on both create and
+   * re-mint responses when the tenant has anonymous sessions enabled.
+   *
+   * Use this — not `expiresAt` — to drive cookie `Max-Age` or session expiry UI.
+   */
+  sessionExpiresAt: number;
```
nandan-bhat: Two things on this line.

The PR description says this was added as an optional `number`, but there is no `?` here or on `AnonymousTokens`. Required is arguably the better choice, but then the runtime has to be able to guarantee it, which is the missing validation I flagged in `parseTokenResponse`. Required plus unvalidated is the worst of the two options, because the type promises a `number` and can hand back `NaN`. Either add the guard or make the field optional, but not neither.

Second, `sessionExpiresAt` already means something else one package over. `auth0-server-js/src/types.ts:149` has `sessionExpiresAt?: number` on `StateData` for the IPSIE `session_expiry` ceiling, wired through `state/utils.ts` and feeding `calculateMaxAge` in `stateless-state-store.ts`. So we would have two different concepts with the same name in neighbouring packages, both driving a cookie `Max-Age`, and `auth0-server-js` will be holding both at once. Could we call it `sessionTokenExpiresAt` here, or at least call out the difference in this doc block?

packages/auth0-auth-js/src/anonymous-session/anonymous-session-client.ts
```diff
@@ -33,6 +33,7 @@ function parseTokenResponse(apiResponse: AnonymousTokenApiResponse): AnonymousTo
     expiresAt: now + expiresIn,
     scope: apiResponse.scope,
     sessionToken: apiResponse.session_token,
+    sessionExpiresAt: now + apiResponse.session_expires_in,
```
nandan-bhat, line 36: This is the only field in this function that is not validated. `access_token` is checked on line 23 and `expires_in` on line 27, both throwing `server_error`. Here a missing field silently gives `now + undefined`, which is `NaN`.

It is not theoretical either. `session_expires_in` is not guaranteed to be in the response. On the renewal path, if the session token being presented carries no expiry of its own, the field is left out of the response body altogether rather than sent as `0` or `null`, so we have to cope with it being absent.

I tried it on a tree with this PR merged. `sessionExpiresAt` comes back as `NaN`, `typeof` is still `number`, and the cookie `Max-Age` derived from it is `NaN`. Nothing throws and nothing logs, the bad value just gets stored and handed to the caller.

Can we guard it the same way as `expires_in`?

packages/auth0-auth-js/src/anonymous-session/types.ts
nandan-bhat, line 53: Two things on this line.

The PR description says this was added as an optional `number`, but there is no `?` here or on `AnonymousTokens`. Required is arguably the better choice, but then the runtime has to be able to guarantee it, which is the missing validation I flagged in `parseTokenResponse`. Required plus unvalidated is the worst of the two options, because the type promises a `number` and can hand back `NaN`. Either add the guard or make the field optional, but not neither.

Second, `sessionExpiresAt` already means something else one package over. `auth0-server-js/src/types.ts:149` has `sessionExpiresAt?: number` on `StateData` for the IPSIE `session_expiry` ceiling, wired through `state/utils.ts` and feeding `calculateMaxAge` in `stateless-state-store.ts`. So we end up with two different concepts under the same name in neighbouring packages, both driving a cookie `Max-Age`, and `auth0-server-js` will be holding both at once. Could we call it `sessionTokenExpiresAt` here, or at least call out the difference in this doc block?

yogeshchoudhary147:

PR #253, opened 2026-08-20
fix(auth0-auth-js): narrow metadata type to Record<string, string>

yogeshchoudhary147:
## What

`CreateAnonymousSessionOptions.metadata` and `AnonymousSessionClaims.metadata` were typed as `Record<string, unknown>`. The platform only accepts string values (`additionalProperties: { type: 'string' }`), so passing non-string values typechecked cleanly but always failed at runtime with a `400 invalid_request`.

## Changes

- Both `metadata` fields narrowed from `Record<string, unknown>` to `Record<string, string>`

## Notes

This is technically a breaking change for callers passing non-string metadata values, but those callers are already broken at runtime. The type change moves the failure from runtime to compile time.

`auth0-server-js` already enforces `Record<string, string>` on its own surface as a workaround — this fixes it in the right place.

### DIFF
--- a/packages/auth0-auth-js/src/anonymous-session/types.ts
+++ b/packages/auth0-auth-js/src/anonymous-session/types.ts
@@ -97,7 +97,7 @@ export interface AnonymousSessionClaims {
   /** Unix timestamp when the anonymous session was created. */
   created_at: number;
   /** Up to 1KB of key-value metadata set at session creation time. */
-  metadata?: Record<string, unknown>;
+  metadata?: Record<string, string>;
 }
 
 /**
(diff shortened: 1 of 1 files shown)


packages/auth0-auth-js/src/anonymous-session/types.ts
```diff
@@ -119,7 +119,7 @@ export interface CreateAnonymousSessionOptions {
    * Up to 1KB of arbitrary key-value metadata to attach to the anonymous session.
    * Set once at creation time — cannot be changed after the session is created.
    */
-  metadata?: Record<string, unknown>;
+  metadata?: Record<string, string>;
```
nandan-bhat, line 122: One thing we are implicitly choosing not to do here, just so it is a deliberate call. There is still no runtime check on the values, so a plain JavaScript caller, or a TypeScript one holding an `any`, can pass `{count: 1}` and get a `400` back from Auth0 instead of a clear SDK error.

Probably fine to leave that to the API, but better as a decision than an oversight.

yogeshchoudhary147:

PR #254, opened 2026-08-20
fix(auth0-auth-js): remove metadata_too_large from error code union

yogeshchoudhary147:
## What

`AnonymousSessionErrorCode` listed `metadata_too_large` under a doc block saying it maps directly to Auth0 API error responses. The platform never sends it — it is an internal validation reason that is dropped before the response is written.

What actually comes back when metadata is too large:
```json
{ "error": "invalid_request", "error_description": "metadata exceeds the maximum allowed size" }
```

Any code checking `error.code === 'metadata_too_large'` is a branch that never runs. The `| string` tail in the union means the compiler does not warn about it.

## Changes

- `errors.ts` — removed `metadata_too_large` from `AnonymousSessionErrorCode`; updated the `invalid_request` doc entry to mention the metadata size case
- `anonymous-session-client.spec.ts` — updated the MSW mock and test assertion from `metadata_too_large` to `invalid_request` to match what the platform actually sends

### DIFF
--- a/packages/auth0-auth-js/src/anonymous-session/errors.ts
+++ b/packages/auth0-auth-js/src/anonymous-session/errors.ts
@@ -13,19 +13,18 @@ export interface AnonymousSessionApiErrorResponse {
  * Codes that map directly to Auth0 API error responses:
  * - `session_expired` — The session token has expired.
  * - `invalid_session_token` — The session token is malformed.
- * - `metadata_too_large` — The metadata payload at session creation exceeded 1 KB.
  * - `unauthorized_client` — The client ID is not enabled for anonymous sessions.
  * - `feature_not_enabled` — The tenant-level anonymous sessions flag is off.
  * - `invalid_client` — Client authentication failed.
  * - `invalid_target` — The resource server is not enabled for anonymous access.
  * - `invalid_request` — The request was malformed or the resource server is not enabled.
+ *   Also returned when the metadata payload exceeds the size limit (~1 KB, JSON-serialized).
  * - `invalid_scope` — The requested scope is not granted to anonymous callers.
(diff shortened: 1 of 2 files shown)


packages/auth0-auth-js/src/anonymous-session/errors.ts
```diff
@@ -13,19 +13,18 @@ export interface AnonymousSessionApiErrorResponse {
  * Codes that map directly to Auth0 API error responses:
  * - `session_expired` — The session token has expired.
  * - `invalid_session_token` — The session token is malformed.
- * - `metadata_too_large` — The metadata payload at session creation exceeded 1 KB.
  * - `unauthorized_client` — The client ID is not enabled for anonymous sessions.
  * - `feature_not_enabled` — The tenant-level anonymous sessions flag is off.
  * - `invalid_client` — Client authentication failed.
  * - `invalid_target` — The resource server is not enabled for anonymous access.
  * - `invalid_request` — The request was malformed or the resource server is not enabled.
+ *   Also returned when the metadata payload exceeds the size limit (~1 KB, JSON-serialized).
```
nandan-bhat: This matches the server exactly, nice catch on the wording. `anonymous_token.js:122` responds with `invalidRequestError(ruleValidation.message)` and the message is `'metadata exceeds the maximum allowed size'`, which is the same string the updated mock now returns.

nandan-bhat, line 21: This matches what the API actually returns, nice catch on the wording. An oversized `metadata` payload comes back as `invalid_request` and the description is `metadata exceeds the maximum allowed size`, which is the same string the updated mock returns.

yogeshchoudhary147:

PR #255, opened 2026-08-20
fix(auth0-auth-js): correct AnonymousSessionClaims shape and unexport AnonymousTokens

yogeshchoudhary147:
## What

Two issues with exported types that no public method returns.

### `AnonymousSessionClaims` — wrong field names

The interface was kept for forward compatibility (session tokens are currently opaque JWEs). The shape was wrong against the actual platform payload:

| Before | After |
|---|---|
| `session_id: string` | removed — does not exist |
| `created_at: number` | removed — does not exist |
| — | `aud: string \| string[]` added |
| — | `iat: number` added |
| — | `exp: number` added |
| — | `sid: string` added |
| `metadata?: Record<string, unknown>` | `metadata?: Record<string, string>` |

Correct shape taken from `SessionTokenService.encryptSessionToken`: `{ iss, sub, aud, iat, exp, sid, metadata }`.

### `AnonymousTokens` — removed from public export

`AnonymousTokens` was exported from `index.ts` but no public method returns it. It is an internal type used between `#postAnonymousToken` and `createSession`/`#mintToken`. Removed from the barrel export; the type itself stays in `types.ts` for internal use.

### DIFF
--- a/packages/auth0-auth-js/src/anonymous-session/index.ts
+++ b/packages/auth0-auth-js/src/anonymous-session/index.ts
@@ -3,7 +3,6 @@ export { AnonymousSessionError } from './errors.js';
 export type { AnonymousSessionErrorCode } from './errors.js';
 export type {
   AnonymousSession,
-  AnonymousTokens,
   AnonymousSessionClaims,
   CreateAnonymousSessionOptions,
   GetAnonymousTokenSilentlyOptions,
--- a/packages/auth0-auth-js/src/anonymous-session/types.ts
+++ b/packages/auth0-auth-js/src/anonymous-session/types.ts
@@ -90,14 +90,18 @@ export interface AnonymousTokens {
 export interface AnonymousSessionClaims {
   /** Issuer of the token. */
   iss: string;
-  /** Subject — the anonymous identity, e.g. `anon@abc123`. */
+  /** Subject — the anonymous identity, e.g. `anon|<id>`. */
   sub: string;
-  /** Unique identifier for the anonymous session. */
-  session_id: string;
-  /** Unix timestamp when the anonymous session was created. */
-  created_at: number;
+  /** Audience — the tenant issuer URL. */
+  aud: string | string[];
+  /** Issued-at timestamp (seconds since epoch). */
+  iat: number;
+  /** Expiry timestamp (seconds since epoch). */
+  exp: number;
+  /** Session ID. */
+  sid: string;
   /** Up to 1KB of key-value metadata set at session creation time. */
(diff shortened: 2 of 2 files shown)


packages/auth0-auth-js/src/anonymous-session/index.ts
```diff
@@ -3,7 +3,6 @@ export { AnonymousSessionError } from './errors.js';
 export type { AnonymousSessionErrorCode } from './errors.js';
 export type {
   AnonymousSession,
-  AnonymousTokens,
```
nandan-bhat: Unexporting this is the right call, `AnonymousTokens` is only referenced inside `anonymous-session-client.ts`.

Since it is becoming internal, this is also the cheap moment to drop `expiresIn` from it (`types.ts:65`). `parseTokenResponse` sets that field and nothing ever reads it, both `createSession` and `#mintToken` return `expiresAt` only. Removing it while the type is still public would be a breaking change, removing it in the same PR that makes it internal costs nothing.

packages/auth0-auth-js/src/anonymous-session/types.ts
```diff
@@ -90,14 +90,18 @@ export interface AnonymousTokens {
 export interface AnonymousSessionClaims {
   /** Issuer of the token. */
   iss: string;
-  /** Subject — the anonymous identity, e.g. `anon@abc123`. */
+  /** Subject — the anonymous identity, e.g. `anon|<id>`. */
```
nandan-bhat, line 93: This one goes the wrong way. Anonymous subjects use `@`, not the usual `provider|id` form. The value is `anon@` followed by a UUID, so it reads like `anon@3f2b1c4e-...` and never like `anon|something`. I decoded an anonymous access token on the test tenant to be sure.

So `anon@abc123` was already correct. Can we make it `anon@<uuid>` rather than `anon|<id>`? Wrong examples have a habit of ending up in docs and in `sub` parsing code.

packages/auth0-auth-js/src/anonymous-session/index.ts
nandan-bhat, line 6: Unexporting this is the right call, `AnonymousTokens` is only referenced inside `anonymous-session-client.ts`.

Since it is becoming internal, this is also the cheap moment to drop `expiresIn` from it (`types.ts:65`). `parseTokenResponse` sets that field and nothing ever reads it, both `createSession` and `#mintToken` return `expiresAt` only. Removing it while the type is still public would be a breaking change, doing it in the same PR that makes it internal costs nothing.

yogeshchoudhary147:

PR #260, opened 2026-08-26
fix(auth0-auth-js): anonymous session post-review fixes

yogeshchoudhary147:
## Summary

Consolidates all post-review fixes for the anonymous session feature introduced in #240.

- **Remove `client_secret` from logout request body** — the anonymous logout endpoint does not accept body-level client credentials; authentication is handled by the `auth0_anon` cookie
- **`private_key_jwt` client authentication** — anonymous token requests now support `client_secret_post`, `private_key_jwt` (PEM or `CryptoKey`), mTLS, and public clients, matching the auth method resolution order of the passkey and passwordless sub-clients
- **`useMtls` wired through to `AnonymousSessionClient`** — prevents sending `client_secret` alongside an mTLS certificate, which Auth0 rejects with `400 Multiple authentication methods provided`
- **Surface `session_expires_in` as `sessionTokenExpiresAt`** — exposes the session token lifetime on `AnonymousSession`; named `sessionTokenExpiresAt` to avoid a clash with the existing `sessionExpiresAt` field on `StateData` in `auth0-server-js`; optional to handle responses where the field is absent
- **Surface `sessionReplaced` on `getAccessToken`** — when a session token is expired or invalid and a fresh identity is silently created, `sessionReplaced: true` is set on the returned session so callers know the previous `sub`, metadata, and any associated state are gone; `false` on a normal renewal
- **Narrow `metadata` to `Record<string, string>`** — aligns the TypeScript type with what the anonymous token endpoint actually accepts
- **Correct `metadata_too_large` error code** — Auth0 returns `invalid_request` for oversized metadata, not a custom `metadata_too_large` code
- **Fix `AnonymousSessionClaims` shape and unexport `AnonymousTokens`** — correct field names (`sid`, `aud`, `iat`, `exp`), fix `sub` example format (`anon@<uuid>`), drop unused `expiresIn` field from the now-internal `AnonymousTokens` type
- **Rename `getTokenSilently` to `getAccessToken`** — the old name was borrowed from `spa-js` silent iframe semantics which do not apply here; `auth0-server-js` already exposes this as `getAccessToken`
- **Document metadata + existing cookie conflict in `createSession`** — in a browser, if an `auth0_anon` cookie is already present when `createSession({ metadata })` is called, Auth0 rejects with `invalid_request`; documented with recovery steps (platform fix tracked separately)

## Test plan

- [ ] `npx turbo run test --filter=@auth0/auth0-auth-js` passes (540 tests)
- [ ] Confirm `useMtls`, `clientAssertionSigningKey`, and `clientAssertionSigningAlg` all flow through to `AnonymousSessionClient`
- [ ] Verify `sessionTokenExpiresAt` is present on `AnonymousSession` and `undefined` when `session_expires_in` is absent
- [ ] Verify `sessionReplaced: true` when `getAccessToken` is called with an expired/invalid session token
- [ ] Verify `sessionReplaced: false` on a normal renewal and absent when called without a `sessionToken`

## Summary by CodeRabbit

* **New Features**
  * Added support for anonymous-session authentication using client secrets, private-key JWTs, public clients, and mutual TLS.
  * Added session-token expiration details and indicators when expired or invalid sessions are replaced.
  * Added clearer token expiration timestamps and standardized session claims.
* **Breaking Changes**
  * Renamed `getTokenSilently` to `getAccessToken`.
  * Anonymous-session metadata now accepts string values and oversized metadata returns `invalid_request`.
* **Documentation**
  * Updated anonymous-session guidance for browser cookies and server-side logout.

### DIFF
--- a/packages/auth0-auth-js/src/anonymous-session/anonymous-session-client.spec.ts
+++ b/packages/auth0-auth-js/src/anonymous-session/anonymous-session-client.spec.ts
@@ -1,8 +1,23 @@
 import { expect, test, describe, beforeAll, afterAll, afterEach, vi } from 'vitest';
 import { setupServer } from 'msw/node';
 import { http, HttpResponse } from 'msw';
+import { decodeJwt, decodeProtectedHeader } from 'jose';
 import { AnonymousSessionClient } from './anonymous-session-client.js';
 import { AnonymousSessionError } from './errors.js';
+
+const exportPrivateKeyToPem = async (privateKey: CryptoKey): Promise<string> => {
+  const pkcs8 = await crypto.subtle.exportKey('pkcs8', privateKey);
+  const keyBase64 = Buffer.from(pkcs8).toString('base64');
+  const keyLines = keyBase64.match(/.{1,64}/g) ?? [keyBase64];
+  return `-----BEGIN PRIVATE KEY-----\n${keyLines.join('\n')}\n-----END PRIVATE KEY-----`;
+};
+
+const generateRsaKeyPair = () =>
+  crypto.subtle.generateKey(
+    { name: 'RSASSA-PKCS1-v1_5', modulusLength: 2048, publicExponent: new Uint8Array([1, 0, 1]), hash: { name: 'SHA-256' } },
+    true,
+    ['sign', 'verify']
+  ) as Promise<CryptoKeyPair>;
 
 const domain = 'auth0.local';
 const clientId = 'test-client-id';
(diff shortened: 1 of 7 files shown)


nandan-bhat:
Remove the internal references (`A1`, `A2`...etc) from the PR descriptions.

yogeshchoudhary147:
Closing in favour of #269 which combines the original anonymous sessions feature (#240) with all these fixes in a single PR.

PR #263, opened 2026-08-26
feat(auth0-server-js): expose getUserInfo delegating to auth-js

tusharpandey13:
### Description

Adds `getUserInfo()` to `ServerClient`, retrieving a user's OIDC profile claims from the `/userinfo` endpoint for an access token you supply.

```ts
const userInfo = await serverClient.getUserInfo({
  accessToken: '<access_token>',
});

console.log(userInfo.sub, userInfo.email, userInfo.name);
```

The access token is passed explicitly. `getUserInfo()` does not read the token from the session and does not trigger a refresh. Pass an optional `expectedSubject` to assert the returned `sub` matches a known value; a mismatch throws `UserInfoError`.

### What's included

- `ServerClient.getUserInfo(options, storeOptions?)`. In resolver (multi-tenant) mode, `storeOptions` is used to resolve the correct domain before the request.
- Re-exports of `GetUserInfoOptions`, `UserInfoResponse`, and `UserInfoError` from `@auth0/auth0-server-js`, so you do not need to import them from `@auth0/auth0-auth-js`.

### Access token requirements

The access token must be one that `/userinfo` accepts:

- **Without Multi-Resource Refresh Tokens (MRRT):** use a default OIDC access token, obtained without an explicit `audience`.
- **With MRRT:** access tokens are audience-bound, so request the userinfo endpoint as the audience (for example, `https://<your-domain>/userinfo`) when obtaining the token. A token bound to a different audience is rejected by `/userinfo`.

See `EXAMPLES.md` for full usage, including resolver mode and error handling.

### DIFF
--- a/packages/auth0-server-js/EXAMPLES.md
+++ b/packages/auth0-server-js/EXAMPLES.md
@@ -48,20 +48,22 @@
   - [Performing a delegation exchange without a session](#performing-a-delegation-exchange-without-a-session)
   - [Using actor tokens for delegation](#using-actor-tokens-for-delegation)
   - [Passing `StoreOptions`](#passing-storeoptions-6)
+- [Retrieving User Information](#retrieving-user-information)
+  - [Passing `StoreOptions`](#passing-storeoptions-7)
 - [Impersonation via Session Transfer](#impersonation-via-session-transfer)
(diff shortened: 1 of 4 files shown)


Piyush-85 (approved):
LGTM. Clean pass-through with resolver-mode support, correct async pattern, good e2e test coverage. Re-exports and EXAMPLES.md are consistent with auth-js.

PR #269, opened 2026-09-02
feat(auth0-auth-js): add anonymous sessions support

yogeshchoudhary147:
## Summary

- Re-lands anonymous sessions support (originally #240, reverted in #268) with all post-review fixes included
- Adds `AnonymousSessionClient` for creating, managing, and exchanging anonymous sessions
- Includes fixes for `private_key_jwt` auth, mTLS support, correct types/claims shape, and `getAccessToken` surface

## Commits

- `feat`: add anonymous sessions support (original #240)
- `fix`: remove client_secret from anonymous session logout body
- `fix`: support private_key_jwt client authentication
- `fix`: add useMtls support to anonymous session client auth
- `fix`: narrow metadata type to `Record<string, string>`
- `fix`: remove `metadata_too_large` from error code union
- `fix`: correct `AnonymousSessionClaims` shape and unexport `AnonymousTokens`
- `fix`: correct sub example and drop dead `expiresIn` from `AnonymousTokens`
- `fix`: surface `session_expires_in` as `sessionTokenExpiresAt`
- `fix`: rename `getTokenSilently` to `getAccessToken` on `AnonymousSessionClient`
- `docs`: document metadata + existing cookie conflict in `createSession`
- `fix`: surface `sessionReplaced` on `getAccessToken` when session is silently replaced

## Test plan

- [x] All existing tests pass
- [x] Anonymous session creation, token exchange, and logout flows work end-to-end
- [ ] `private_key_jwt` and mTLS client auth variants verified

🤖 Generated with [Claude Code](https://claude.com/claude-code)

## Summary by CodeRabbit

* **New Features**
  * Added anonymous sessions for creating, renewing, and ending sessions.
  * Exposed anonymous session access through the main authentication client.
  * Added support for public clients, client secrets, private key authentication, and mutual TLS.
  * Sessions can silently renew when tokens expire or become invalid.
  * Added configurable audiences, scopes, and session metadata.
  * Added structured errors with machine-readable error codes.

* **Tests**
  * Added comprehensive coverage for session lifecycle, authentication, renewal, expiration, metadata, and error handling.

### DIFF
--- a/packages/auth0-auth-js/src/anonymous-session/anonymous-session-client.ts
+++ b/packages/auth0-auth-js/src/anonymous-session/anonymous-session-client.ts
@@ -6,7 +6,7 @@ import type {
   AnonymousTokens,
   AnonymousTokenApiResponse,
   CreateAnonymousSessionOptions,
-  GetAnonymousTokenSilentlyOptions,
+  GetAnonymousAccessTokenOptions,
 } from './types.js';
 
 /**
@@ -59,7 +59,7 @@ async function parseErrorResponse(response: Response): Promise<AnonymousSessionA
  * Client for anonymous session operations.
  *
  * Provides low-level HTTP calls to the Auth0 anonymous session endpoints as well
- * as a higher-level {@link getTokenSilently} method that implements the renewal
+ * as a higher-level {@link getAccessToken} method that implements the renewal
  * logic: if the access token is expired it re-mints it from the session token;
  * if the session token itself has expired a fresh anonymous session is created
  * silently (any metadata previously set on the session is lost at that point).
@@ -81,7 +81,7 @@ async function parseErrorResponse(response: Response): Promise<AnonymousSessionA
  * });
  *
  * // Later, get a valid access token (renews automatically if expired)
- * const updatedSession = await authClient.anonymous.getTokenSilently({
+ * const updatedSession = await authClient.anonymous.getAccessToken({
  *   sessionToken: session.sessionToken,
  *   audience: 'https://api.example.com',
  * });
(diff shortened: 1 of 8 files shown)


packages/auth0-auth-js/src/anonymous-session/anonymous-session-client.ts
```diff
@@ -0,0 +1,369 @@
...
+    };
+
+    const response = await this.#customFetch(url, {
+      method: 'POST',
+      headers: {
+        'Content-Type': 'application/json',
+      },
+      credentials: 'include',
+      body: JSON.stringify(body),
+    });
```
cschetan77, line 312: `#postAnonymousToken` sets redirect: 'error' but `logout()` doesn't. Worth adding here for consistency, both methods POST to anonymous session endpoints with credentials: 'include', so a reader seeing the two fetch calls side by side will reasonably ask why they differ.

The concrete security risk is low (the logout body is just `{ client_id }`, and cookies won't follow cross-origin redirects anyway), but keeping the options identical makes the intent clear and means the protection is already in place if client auth is ever added to logout later.

packages/auth0-auth-js/src/anonymous-session/utils.ts
```diff
@@ -0,0 +1,73 @@
...
+ * client (no auth options configured): the anonymous token endpoint does not
+ * require client authentication for public clients.
+ *
+ * @internal
+ */
+export async function buildClientAuthBody(
+  options: ClientAuthOptions,
+  clientId: string,
+  domain: string
+): Promise<Record<string, string>> {
```
cschetan77, line 34: The buildClientAuthBody implementation here is nearly identical to the one in `passwordless/utils.ts`, 
same constants, same imports, same JWT signing block, same claim shape. The only meaningful difference is the public client path: passwordless throws `MissingClientAuthError`, anonymous session returns `{}` (which is correct since the anonymous token endpoint supports public clients).

Worth considering extracting the shared JWT signing logic into a common utility (e.g. src/client-auth.ts) that both can import from. Any future change to the assertion format, audience, or claim shape would otherwise need to be applied in two places. Not a blocker for this PR, leaving it to your judgement.

packages/auth0-auth-js/src/anonymous-session/anonymous-session-client.ts
yogeshchoudhary147: Good catch. Risk is low today since body only has `client_id` but good to have it consistent.

packages/auth0-auth-js/src/anonymous-session/utils.ts
yogeshchoudhary147: Makes sense, but the two functions behave differently for public clients by design so keeping them separate is cleaner than adding a flag just to handle that one case.

cschetan77 (dismissed):

cschetan77 (approved):

PR #270, opened 2026-09-03
feat(auth0-server-js): expose Experiment Center override params on AuthorizationParameters

cschetan77:
## Summary

- Adds `experiment_id`, `variation_id`, and `segment_id` as named, typed `string` fields on `AuthorizationParameters`
- Params are forwarded to `/authorize` via the existing `authorizationParams` spread — no runtime logic changes needed
- Adds unit tests verifying the params appear on the `/authorize` URL (with and without `segment_id`)
- Adds a `### Using Experiment Center` section to `EXAMPLES.md` with usage examples

## What are these params?

[Experiment Center](https://auth0.com/docs/customize/experiment-center/overview) is Auth0's A/B testing platform. These three params let developers force a specific experiment variation on a `/authorize` request, bypassing the server-side deterministic assignment.

## Test plan

- [ ] `npx vitest run packages/auth0-server-js/src/server-client.spec.ts` — all tests pass
- [ ] Verify `experiment_id`, `variation_id`, `segment_id` appear in `AuthorizationParameters` typings with correct JSDoc
- [ ] Review `EXAMPLES.md` — new section renders correctly

## Summary by CodeRabbit

* **New Features**
  * Added optional Experiment Center parameters to interactive login authorization requests, allowing applications to select a specific experiment variation and optionally target a segment.
  * Parameters can be provided per login to avoid applying overrides to every login.

* **Documentation**
  * Added usage guidance and a table-of-contents entry for configuring Experiment Center overrides during interactive login.
  * Documented that Experiment Center is an Enterprise feature and that overrides should be supplied per login.

### DIFF
--- a/packages/auth0-server-js/EXAMPLES.md
+++ b/packages/auth0-server-js/EXAMPLES.md
@@ -22,6 +22,7 @@
   - [Passing `authorizationParams`](#passing-authorization-params)
   - [Passing `appState` to track state during login](#passing-appstate-to-track-state-during-login)
   - [Logging in to an Organization](#logging-in-to-an-organization)
+  - [Using Experiment Center](#using-experiment-center)
   - [Using Pushed Authorization Requests](#using-pushed-authorization-requests)
   - [Using Pushed Authorization Requests and Rich Authorization Requests](#using-pushed-authorization-requests-and-rich-authorization-requests)
   - [Passing `StoreOptions`](#passing-storeoptions)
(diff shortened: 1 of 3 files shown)


nandan-bhat:
Should we EC params to the `AuthorizationParameters` interface in `auth0-auth-js` as-well ?
I am talking about this: https://github.com/auth0/auth0-auth-js/blob/main/packages/auth0-auth-js/src/types.ts#L136C18-L136C41


If at all we are making this change, it has to go in a separate PR. But calling it out here.

packages/auth0-server-js/EXAMPLES.md
```diff
@@ -712,6 +713,33 @@ try {
 
 See the [Auth0 Organizations documentation](https://auth0.com/docs/manage-users/organizations) for setup and concepts.
 
+### Using Experiment Center
+
+Pass `experiment_id` and `variation_id` via `authorizationParams` to force a user into a specific [Experiment Center](https://auth0.com/docs/customize/experiment-center/overview) variation for the login request, bypassing the server-side deterministic assignment. Both IDs are obtained from your Auth0 Dashboard or the Management API.
+
+```ts
+await serverClient.startInteractiveLogin({
+  authorizationParams: {
```
nandan-bhat, line 722: Could we add a short note that these should be passed per login through `startInteractiveLogin`, and not set on the client at construction time ? If they go into the client level `authorizationParams`, every login gets pinned to the same variation, which quietly cancels out the A/B test."

But if that is what the customer wants, let them do it. It's better to add that information.

packages/auth0-server-js/src/server-client.spec.ts
```diff
@@ -721,6 +721,57 @@ test('startInteractiveLogin - should build the authorization url with custom par
   expect(url.searchParams.size).toBe(7);
 });
 
+test('startInteractiveLogin - should include experiment override params on the authorization url', async () => {
```
nandan-bhat, line 724: These tests set the params and check they reach the `/authorize` URL. That already worked before this PR through the index signature (`[key: string]: unknown`), so they would still pass even if the new typed fields were removed.

So they cover the forwarding and not the new types. Fine to keep, just noting that the types are not really exercised here. A type level check would be the only way to catch a regression on them."

packages/auth0-server-js/EXAMPLES.md
cschetan77: yes makes sense, I've added a `>Important` note just below this section calling out that clearly.

cschetan77:
> Should we EC params to the `AuthorizationParameters` interface in `auth0-auth-js` as-well ? I am talking about this: https://github.com/auth0/auth0-auth-js/blob/main/packages/auth0-auth-js/src/types.ts#L136C18-L136C41
> 
> If at all we are making this change, it has to go in a separate PR. But calling it out here.

I intentionally kept EC params out of `auth0-auth-js's` AuthorizationParameters. My though level is, that interface is a minimal protocol-level contract (scope, audience, redirect_uri) and it's the foundation that any OAuth client could build on. EC params belong at the framework SDK layer, which is where developers interact with auth flows. This keeps the auth0-auth-js completely minimalistic, generic and bloat free. 
Let me know if you think otherwise.

nandan-bhat (approved):

PR #283, opened 2026-09-17
feat(auth0-auth-js): add Session Transfer Ticket support to AnonymousSessionClient

yogeshchoudhary147:
## Summary

- Adds `mintTransferToken(sessionToken)` to `AnonymousSessionClient` — calls `POST /anonymous/token` with `audience: "urn:auth0:anon_transfer"` and returns the resulting 30s single-use JWE. Returns `null` on any failure so callers are never blocked from proceeding with login.
- Adds `anon_transfer_token?: string` to `AuthorizationParameters` as an explicit typed field for upper-layer SDKs to append the ticket to the `/authorize` URL.

## Test plan

- [ ] All 573 `auth0-auth-js` unit tests pass

## Summary by CodeRabbit

* **New Features**
  * Added support for transferring an anonymous session into an authenticated login flow.
  * Added a method to obtain a short-lived transfer token for connecting anonymous activity with a user’s authenticated identity.
  * Authorization requests can now include an optional anonymous-session transfer token.
  * Transfer-token failures are handled gracefully, allowing authentication to continue without blocking the login experience.


packages/auth0-auth-js/src/types.ts
```diff
@@ -147,6 +147,16 @@ export interface AuthorizationParameters {
...
+  /**
+   * A short-lived (30 s) single-use JWE that links an active anonymous session to the
+   * login flow. Minted via `POST /anonymous/token` with
+   * `audience: "urn:auth0:anon_transfer"` and appended to the `/authorize` URL so the
+   * platform can associate the anonymous identity with the authenticated user.
+   *
+   * SDKs that support anonymous sessions set this automatically — you do not need to
+   * pass it manually.
+   */
+  anon_transfer_token?: string;
```
cschetan77: IMO `anon_transfer_token` shouldn't live on AuthorizationParameters in `auth0-auth-js`. The core SDK's `AuthorizationParameters` should be OAuth/OIDC primitives: scope, audience, redirect_uri - parameters that a caller might legitimately configure. `anon_transfer_token` is an internal SDK-to-platform detail that:

- Framework SDKs (server-js, spa-js) inject silently and automatically, it's never part of the caller-facing API in either SDK
- The JSDoc itself acknowledges this: "SDKs that support anonymous sessions set this automatically —you do not need to pass it manually"
- Making it an explicit typed field on the core interface exposes it to every consumer of auth0-auth-js as a legitimate thing to set, which creates a confusing and misleading API surface

The [key: string]: unknown index signature on AuthorizationParameters already exists for exactly this kind of internal SDK parameter.

This keeps the core auth-js SDK clean and minimilistic.

packages/auth0-auth-js/src/anonymous-session/anonymous-session-client.ts
```diff
@@ -321,6 +321,68 @@ export class AnonymousSessionClient {
...
+        this.#domain
+      );
+      Object.assign(body, authFields);
+
+      const response = await this.#customFetch(url, {
+        method: 'POST',
+        headers: { 'Content-Type': 'application/json' },
+        credentials: 'include',
+        redirect: 'error',
+        body: JSON.stringify(body),
```
cschetan77: Passing  `credentials: 'include'` seems to be a problem here. 

First, this method is specifically for exchanging the session token with a transfer ticket, including a cookie doesn't seem to serve any purpose other than a security risk. 

Second, the Auth0 platform's session token resolution prioritises the cookie over the request body,  so in a browser context, if an `auth0_anon` cookie is present it silently overrides the session_token we pass in the body. In the happy path both refer to the same session, but if they've drifted (e.g. localStorage cleared but cookie not, or cookie from a previous session) the transfer ticket is minted for the cookie's session, not the one the caller passed. The SDK has no way to detect or signal this.

Unlike `createSession` and `getAccessToken`, where the cookie being the source of truth is intentional, `mintTransferToken` takes an explicit session token as its input, implying the caller knows exactly which session to transfer. credentials: 'include' undermines that contract.

Removing it is a safe bet - 
- Server-side (RWA / Node.js): credentials: 'include' is a no-op on native fetch, no change in behaviour.
- Browser (SPA): session is resolved from the explicit session_token in the body, which is the intent.

